PR #29, opened 2019-09-12
Kseniia's solution

KseniiaMazan:
[DEMO LINK](https://KseniiaMazan.github.io/react_tabs/)

### DIFF
--- a/src/Components/Tabs/Tabs.js
+++ b/src/Components/Tabs/Tabs.js
@@ -9,37 +9,36 @@ class Tabs extends React.Component {
 
     this.state = {
       activeTab: 0,
-      isActive: 0,
     };
   }
 
   onTabSelected = (index) => {
     this.setState({
       activeTab: index,
-      isActive: index,
     });
   }
 
   render() {
     const { tabs } = this.props;
+    const { activeTab } = this.state;
 
     return (
       <>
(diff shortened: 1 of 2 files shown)


src/Components/Tabs/Tabs.js
```diff
@@ -0,0 +1,51 @@
...
+import { tabsPropTypes } from '../PropTypes/PropTypes';
+import './Tabs.css';
+
+class Tabs extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      activeTab: 0,
+      isActive: 0,
```
irenhh: `isActive` name assumes that you store false or true value inside

src/Components/Tabs/Tabs.js
```diff
@@ -0,0 +1,51 @@
...
+
+  onTabSelected = (index) => {
+    this.setState({
+      activeTab: index,
+      isActive: index,
+    });
+  }
+
+  render() {
+    const { tabs } = this.props;
```
irenhh: destructure all props

src/Components/Tabs/Tabs.js
irenhh: why do you need two similar variables?

PR #37, opened 2019-09-13
Vika Tsykun solution

ViktoriiaTsykun:
[LINK](https://viktoriiatsykun.github.io/react_tabs/)


src/components/Tab/Tab.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import './Tab.css';
+import { TabPropTypes } from '../../proptypes/proptypes';
+
+const Tab = ({
+  tab, onTabSelected, index, activeTab,
+}) => {
+  const className = i => (
```
irenhh: `getClassName` is a better option since it's a function and you have to name it properly

PR #43, opened 2019-09-14
Solution

marchukmiroslava88:
[DEMO](https://marchukmiroslava88.github.io/react_tabs/)


src/Tab/Tab.js
```diff
@@ -0,0 +1,10 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Tab = props => <div className="tab-pane">{props.children}</div>;
```
danheim:
```suggestion
const Tab = ({ children }) => <div className="tab-pane">{children}</div>;
```
why not

PR #52, opened 2019-09-23
Develop

ibezkorovayniy:
https://ibezkorovayniy.github.io/react_tabs/

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.css';
+import Tabs from './components/Tabs/Tabs';
 
 class App extends React.Component {
   state = {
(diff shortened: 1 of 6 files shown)


src/App.js
```diff
@@ -12,11 +13,19 @@ class App extends React.Component {
 
   render() {
     const { tabs } = this.state;
+    let index = 0;
+    const tabsWithId = tabs.map(tab => (
```
YuriiHoliuk: do you really want to run this staff each render?

src/components/Tab/Tab.js
```diff
@@ -0,0 +1,23 @@
+import React from 'react';
+
+const Tab = (props) => {
+  const {
+    handleClick, title, activeTab, id,
```
YuriiHoliuk: will be better for Tab to receive prop isActive instead of activeTab

YuriiHoliuk: will be better to name prop onClick instead of handleClick here, but name this function handleClick in parent component

src/components/Tabs/Tabs.js
```diff
@@ -0,0 +1,41 @@
...
+              handleClick={() => this.onTabSelected(tab.id)}
+              title={tab.title}
+              activeTab={activeTab}
+              id={tab.id}
+              key={tab.id}
+            />
+          ))}
+        </div>
+        <div className="tab-body">
+          {tabs[activeTab].content}
```
YuriiHoliuk: you use id as index here

PR #85, opened 2019-12-05
solution

OlehTereshchuk:
- [DEMO LINK](https://OlehTereshchuk.github.io/react_tabs/)


src/Tabs.js
```diff
@@ -0,0 +1,43 @@
...
+    this.state = { index: 0 };
+  }
+
+  switchTab(i) {
+    this.setState({ index: i });
+  }
+
+  render() {
+    const { props: [...arr] } = this.props;
+
```
vkryvytskyy: const { index } = this.state

PR #178, opened 2020-05-03
solution

Vlad-Viddi:
[DEMO LINK](https://vlad-viddi.github.io/react_tabs/)

### DIFF
--- a/src/Tabs/Tabs.js
+++ b/src/Tabs/Tabs.js
@@ -1,0 +1,50 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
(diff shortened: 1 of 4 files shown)


src/Tabs/Tabs.js
```diff
@@ -0,0 +1,50 @@
...
+
+  onTabsSelected = (index) => {
+    this.setState({
+      selectedTab: index,
+    });
+  }
+
+  render() {
+    const { table } = this.props;
+
```
vkryvytskyy, line 17: const { selectedTab } =  this.state;

PR #179, opened 2020-05-03
Game2048

RustamKhananov:
[DEMO LINK](https://RustamKhananov.github.io/react_tabs/)


src/App.js
```diff
@@ -1,24 +1,406 @@
...
+              ));
+            }
+          }
+        }
+      }
+    }
+  }
+
+  handleTouchStart = (event) => {
+    // event.nativeEvent.preventDefault()
```
irenhh, line 323: remove commented code if you don't use it

src/App.js
```diff
@@ -1,24 +1,406 @@
...
+            if (a !== 0 && b !== 0) {
+              return 0;
+            }
+
+            if (a !== 0 && b === 0) {
+              return 1;
+            }
+
+            return 0;
+          });
```
irenhh, line 208: you have the same part for two cases so it would be good to add separate function for this and use it in both places

src/App.js
```diff
@@ -1,24 +1,406 @@
...
+              tempColumnArr[i2] *= 2;
+              if (tempColumnArr[i2] === 2048) {
+                this.playerWins();
+              }
+
+              tempColumnArr.splice(i2 - 1, 1);
+              tempColumnArr.unshift(0);
+              plusZeros += 1;
+            }
+          }
```
irenhh, line 222: the same for this one

PR #213, opened 2020-09-05
solution with linter ignoring

Andrii256:
[DEMO LINK](https://andrii256.github.io/react_tabs/)


rmnkk (approved):
Nice. Bootstrap 👍

src/components/Tabs/Tabs.js
```diff
@@ -0,0 +1,45 @@
...
+
+  return (
+    <div className="Tabs">
+      <div className="nav nav-tabs">
+        {tabs.map((elem, index) => (
+          <li className="nav-item">
+            <a
+              href="#linterPlug"
+              type="button"
+              key={elem.title.split(' ').join('')} // word without spaces
```
irenhh, line 17: what is the problem with spaces?

Andrii256: it is bad key with spaces I think

Andrii256: I can change back, but as for me this is better

irenhh: I don't see any problems in spaces for key, it just have to be unique

irenhh: FYI ⬆️
I'm approving

irenhh (approved):

PR #215, opened 2020-09-06
done

nanccyy:
[DEMO LINK](https://nanccyy.github.io/react_tabs/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.css';
+import Tab from './Tab';
 
 // eslint-disable-next-line no-unused-vars
 const tabs = [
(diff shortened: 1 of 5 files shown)


src/App.js
```diff
@@ -17,8 +18,35 @@ const tabs = [
   },
 ];
 
-const App = () => (
-  <h1>React tabs</h1>
-);
+class App extends React.Component {
+  state = {
+    number: 0,
```
varseniuk, line 23: I think it would be better to give this property another name. Something like 'currentTabIndex', because the name 'number' gives us no certain information about what's in it.
But it's only my opinion. Everything else is 👍

varseniuk (approved):
Nice work 👍

src/App.js
```diff
@@ -17,8 +18,35 @@ const tabs = [
   },
 ];
 
-const App = () => (
-  <h1>React tabs</h1>
-);
+class App extends React.Component {
+  state = {
+    currentTabIndex: 0,
+    content: tabs[0].content,
```
Denchepornyuk, line 24: в state потрібно зберігати тільки індекс активної таби

Denchepornyuk (approved):

PR #340, opened 2021-01-25
solution

troospieler:
[DEMO LINK](https://troospieler.github.io/react_tabs/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,5 +1,6 @@
 import React from 'react';
 import './App.css';
+import { Tabs } from './components/Tabs';
 
 // eslint-disable-next-line no-unused-vars
 const tabs = [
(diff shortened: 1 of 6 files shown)


taniazt (approved):

src/App.js
```diff
@@ -17,8 +18,36 @@ const tabs = [
...
+  }
+
+  onTabSelected = (event) => {
+    const chosenTab = tabs.find(tab => (
+      tab.title === event.target.innerText
+    ));
+
+    this.setState({
+      index: tabs.indexOf(chosenTab),
+    });
```
aliceblunt8, line 33: you can also pass tab.id to this function, so you will only need to save it to the state (no need to use find and indexOf) ;)

src/components/Tabs/Tabs.jsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import './Tabs.css';
+import classNames from 'classnames';
+
+export class Tabs extends React.Component {
+  state = {
+    tabs: this.props.tabs,
+  }
```
aliceblunt8, line 9: no need to save tabs in state, just use them as props

aliceblunt8 (changes requested):

aliceblunt8 (approved):

PR #353, opened 2021-02-01
solution

Alexandr-Petrenko:
[DEMO LINK](https://Alexandr-Petrenko.github.io/react_tabs/)

### DIFF
--- a/src/components/Tab/Tab.jsx
+++ b/src/components/Tab/Tab.jsx
@@ -1,0 +1,50 @@
+
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+export class Tab extends Component {
+  state = {
+    currentTab: this.props.tabs[0].title,
+  }
+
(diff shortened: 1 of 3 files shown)


src/components/Tab/Tab.jsx
```diff
@@ -0,0 +1,50 @@
...
+          {tabs.map(tab => (
+            <button
+              key={tab.title}
+              className={classNames({
+                tabs__btn: true,
+                tabs__btn_active: tab.title === this.state.currentTab,
+              })}
+              type="button"
+              onClick={() => {
+                this.tabToggler(tab);
```
varseniuk, line 30:
```suggestion
                this.tabToggler(tab.title);
```
You can pass here only 'title', not the whole 'tab' object (without 'content' field)
And your 'tabToggler' function should look like this:
```
 tabToggler = (title) => {
    this.setState({ currentTab: title });
  }
```

varseniuk (approved):
Good, but read my small comment 👍

PR #420, opened 2021-05-24
sol

Ihor-Karpyn:
- [DEMO LINK](https://Ihor-Karpyn.github.io/react_tabs/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,19 +1,44 @@
 import React from 'react';
 
 import './App.scss';
+import Tabs from './components/Tabs/Tabs';
 
 /* eslint-disable */
-const tabs = [
-  { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
-  { id: 'tab-2', title: 'Tab 2', content: 'Some text 2' },
-  { id: 'tab-3', title: 'Tab 3', content: 'Some text 3' },
+const tabsFromServer = [
+  {id: 'tab-1', title: 'Tab 1', content: 'Some text 1'},
+  {id: 'tab-2', title: 'Tab 2', content: 'Some text 2'},
+  {id: 'tab-3', title: 'Tab 3', content: 'Some text 3'},
 ];
+
 /* eslint-enable */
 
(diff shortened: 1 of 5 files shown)


src/components/Tabs/Tabs.jsx
```diff
@@ -0,0 +1,54 @@
...
+    const { idSelectedTab, onChangeSelTabs } = this.props;
+
+    if (idSelectedTab !== id) {
+      onChangeSelTabs(id);
+    }
+  }
+
+  render() {
+    const { tabs, idSelectedTab } = this.props;
+    const selectedTab = tabs.find(tab => tab.id === idSelectedTab);
```
IvanGrekov, line 17: We already have such string in the component App. It would be better if we find selectedTab only once

src/components/Tabs/Tabs.jsx
```diff
@@ -0,0 +1,54 @@
+import React from 'react';
+import './Tabs.scss';
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+class Tabs extends React.Component {
+  changeStatusHandler = (id) => {
+    const { idSelectedTab, onChangeSelTabs } = this.props;
```
IvanGrekov, line 8: For this component you can create a function, not a class. It helps you make destructuring of props only one time. Now you make it here and into the render of this component

IvanGrekov (approved):
Good job. But next time try to avoid unnecessary duplicating of code

src/App.js
```diff
@@ -1,19 +1,44 @@
...
 /* eslint-enable */
 
-const App = () => (
-  <div className="App">
-    <h1>Selected tab is&nbsp;Tab 1</h1>
-  </div>
-);
+class App extends React.Component {
+  state = {
+    idSelectedTab: tabsFromServer[0].id,
```
mykola-yarchuk, line 17: keep selected tab here instead of id, that let you take the title from it

mykola-yarchuk (approved):

PR #421, opened 2021-05-27
react_tabs

IvanGrekov:
[DEMO LINK](https://IvanGrekov.github.io/react_tabs/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import { Tabs } from './components/Tabs';
 
 import './App.scss';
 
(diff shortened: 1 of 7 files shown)


src/components/Tabs/Tabs.js
```diff
@@ -0,0 +1,60 @@
...
+            <li
+              className="tabs__switcher"
+              key={tab.id}
+              onClick={() => {
+                onClick(tab.id);
+              }}
+            >
+              {tab.title}
+            </li>
+          );
```
Ihor-Karpyn, line 37: So difficult

Ihor-Karpyn (approved):
Great

src/App.js
```diff
@@ -10,10 +11,36 @@ const tabs = [
...
+    selectedTab: tabs[0],
+  }
+
+  changeSelectedTab = (tabId) => {
+    this.setState(() => {
+      const newSelectedTab = tabs.find(tab => tab.id === tabId);
+
+      return {
+        selectedTab: newSelectedTab,
+      };
```
mykola-yarchuk, line 25: image
if user clicks on the tab that already selected setState shouldn't be called

src/App.js
```diff
@@ -10,10 +11,36 @@ const tabs = [
...
+  render() {
+    const { selectedTab } = this.state;
+
+    return (
+      <div className="App">
+        <h1>{`Selected tab is ${selectedTab.title}`}</h1>
+
+        <Tabs
+          tabsList={tabs}
+          selectedTab={selectedTab}
```
mykola-yarchuk, line 38: pass only id

mykola-yarchuk (changes requested):

src/App.js
IvanGrekov: But in the component Tabs there is element `paragraph`, which receive selectedTab.content

src/components/Tabs/Tabs.js
```diff
@@ -0,0 +1,47 @@
...
+import PropTypes from 'prop-types';
+import classNames from 'classnames';
+
+import './Tabs.scss';
+
+export const Tabs = React.memo(
+  ({
+    tabsList,
+    selectedTabId,
+    onClick,
```
VinnieJ-2k20, line 11: Better not to call props fields the same as existing attributes.

VinnieJ-2k20 (approved):
Good job 👍

PR #433, opened 2021-05-28
Solution

ViPiven:
[DEMO LINK](https://ViPiven.github.io/react_tabs/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,6 +1,7 @@
 import React from 'react';
 
 import './App.scss';
+import Tabs from './components/Tabs';
 
 /* eslint-disable */
 const tabs = [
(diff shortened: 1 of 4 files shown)


ched88xtc (approved):
LGTM!

src/App.js
```diff
@@ -10,10 +11,35 @@ const tabs = [
...
-  </div>
-);
+class App extends React.Component {
+  state = {
+    selectedTab: tabs[0],
+  }
+
+  onTabSelected = (tabId) => {
+    const newTab = tabs.find(tab => tab.id === tabId);
+
```
varseniuk, line 21: As task description says:

> Don't call the callback if the tab was not changed
So you should add the check if clicked tab is === seledted tab

src/App.js
```diff
@@ -10,10 +11,35 @@ const tabs = [
...
+    return (
+      <div className="App">
+        <h1>
+          Selected tab is&nbsp;
+          {this.state.selectedTab.title}
+        </h1>
+
+        <Tabs
+          tabs={tabs}
+          selectedTab={this.state.selectedTab}
```
varseniuk, line 37:
```suggestion
          selectedTabId={this.state.selectedTab.id}
```

> Pass the `selectedTabId` as a prop to the Tabs, the specified tab should be selected if possible (otherwise the first tab is selected)

You are passing the whole `tabs` array. You may pass only `id` and find needed tab there by id

varseniuk (changes requested):
Let's make small changes

src/components/Tabs.jsx
```diff
@@ -0,0 +1,35 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+function Tabs({ tabs, selectedTabId, onTabSelected }) {
+  return (
+    <>
```
aliceblunt8, line 6: remove fragment

aliceblunt8 (approved):

PR #461, opened 2021-05-31
Develop

cheytac200:
[DEMO LINK](https://cheytac200.github.io/react_tabs/)


src/App.jsx
```diff
@@ -0,0 +1,41 @@
...
+  { id: 'tab-1', title: 'Home', content: 'Some text 1' },
+  { id: 'tab-2', title: 'Profile', content: 'Some text 2' },
+  { id: 'tab-3', title: 'Contact', content: 'Some text 3' },
+];
+/* eslint-enable */
+
+class App extends React.Component {
+  state = {
+    selectedTabId: tabs[0].id,
+    selectedTitle: tabs[0].title,
```
VinnieJ-2k20, line 17: Store active tab in state and use any of its fields when necessary.
```suggestion
    selectedTab: tabs[0],
```

src/App.jsx
```diff
@@ -0,0 +1,41 @@
...
+  };
+
+  selectTab = ({ id, title }) => {
+    this.setState({
+      selectedTabId: id,
+      selectedTitle: title,
+    });
+  }
```
VinnieJ-2k20, line 25: If user clicks already selected tab, `setState` should not be called.

src/components/Tabs.jsx
```diff
@@ -0,0 +1,43 @@
...
+
+      {tabs.map((tab) => {
+        if (selectedTabId === tab.id) {
+          return (
+            <p key={tab.id}>{tab.content}</p>
+          );
+        }
+
+        return false;
+      })}
```
VinnieJ-2k20, line 26: Better to use find method to find active tab and display its content here.
The less JS logic is inside `jsx` - the better.

VinnieJ-2k20 (changes requested):
Almost done. 
Some edits needed.

PR #464, opened 2021-05-31
Solution

chumasov:
[DEMO LINK](https://chumasov.github.io/react_tabs/)

### DIFF
--- a/src/components/Tabs.jsx
+++ b/src/components/Tabs.jsx
@@ -1,0 +1,31 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const Tabs = ({ tabs, selectedTabId, onTabSelected }) => (
+  tabs.map(tab => (
+    <button
+      key={tab.id}
+      type="button"
+      className={selectedTabId === tab.id ? 'selected' : ''}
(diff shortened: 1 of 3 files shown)


VinnieJ-2k20 (changes requested):
All looks good, just forgot to render content of active tab.

VinnieJ-2k20 (changes requested):
OK, content is now rendering.
But it should be rendered in the same element for all tabs, like in the example:
Hint - put it outside map method.

src/components/Tabs.jsx
```diff
@@ -17,12 +18,16 @@ export const Tabs = ({ tabs, selectedTabId, onTabSelected }) => (
...
-    </>
-  ))
+      ))}
+    </div>
+    <div className="tabs__content">
+      {tabs.map(tab => (
+        <React.Fragment key={tab.id}>
+          {selectedTabId === tab.id && <p>{tab.content}</p>}
+        </React.Fragment>
+      ))}
```
VinnieJ-2k20, line 28: There is more simple way - use `find` method here.

VinnieJ-2k20 (approved):

PR #465, opened 2021-05-31
Solution

vitaliy-buchinchik:
[DEMO LINK](https://vitaliy-buchinchik.github.io/react_tabs/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,4 +1,5 @@
 import React from 'react';
+import Tabs from './components/Tabs/Tabs';
 
 import './App.scss';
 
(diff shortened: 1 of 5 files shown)


src/App.js
```diff
@@ -10,10 +11,34 @@ const tabs = [
...
+class App extends React.Component {
+  state = {
+    selectedTab: tabs[0],
+  }
+
+  onTabSelected = (tab) => {
+    this.setState({
+      selectedTab: tab,
+    });
+  }
```
VinnieJ-2k20, line 23: If user clicks on already selected tab, `setState` should not be called.

VinnieJ-2k20 (approved):
I'll approve, but try to follow the task guidelines.

src/App.js
vitaliy-buchinchik: Ok, I`ll fix it

PR #466, opened 2021-05-31
Solution

bohdanklius:
- [DEMO LINK](https://bohdanklius.github.io/react_tabs/)

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -2,6 +2,8 @@ import React from 'react';
 
 import './App.scss';
 
+import { Tabs } from './components/Tabs/Tabs';
+
 /* eslint-disable */
 const tabs = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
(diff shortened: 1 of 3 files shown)


src/App.js
```diff
@@ -10,10 +12,34 @@ const tabs = [
...
+class App extends React.Component {
+  state = {
+    selectedTab: tabs[0],
+  }
+
+  onTabSelect = (tab) => {
+    this.setState({
+      selectedTab: tab,
+    });
+  }
```
VinnieJ-2k20, line 24: If user clicks the tab that is already selected, `setState` should not be called.

src/App.js
```diff
@@ -10,10 +12,34 @@ const tabs = [
...
+    return (
+      <div className="App">
+        <h1>
+          Selected tab is
+          {selectedTab.title}
+        </h1>
+        <Tabs
+          tabs={tabs}
+          onTabSelect={this.onTabSelect}
+          selectedTab={selectedTab}
```
VinnieJ-2k20, line 38: Better to pass only id here. You have access to all tabs in `tabs` prop.

src/components/Tabs/Tabs.jsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+export const Tabs = ({ tabs, onTabSelect, selectedTab }) => (
+  <div>
+    <ul>
+      {tabs.map(tab => (
+        <button
+          className={tab.id === selectedTab.id ? 'active' : ''}
```
VinnieJ-2k20, line 9: Better to use `classnames` library here.

VinnieJ-2k20 (changes requested):
Almost done, need to make some edits.

VinnieJ-2k20 (approved):

PR #540, opened 2021-09-09
solution

VeronikaLushchik:
[DEMO LINK](https://VeronikaLushchik.github.io/react_tabs/)


SashaRudenko (approved):

src/App.tsx
```diff
@@ -1,18 +1,46 @@
...
-    <h1>Selected tab is&nbsp;Tab 1</h1>
-    <div>{tabs.length}</div>
-  </div>
-);
+interface State {
+  selectedTab: Tab
+}
+
+class App extends React.Component<{}, State> {
+  state = {
```
Elena-brv, line 18:
```suggestion
  state: State = {
```

src/components/Tabs.tsx
```diff
@@ -0,0 +1,40 @@
...
+              onClick={() => onTabSelected(tab.id)}
+              className={classNames(
+                'nav-link',
+                {
+                  active: tab.id === selectedTabId,
+                },
+              )}
+            >
+              {tab.title}
+            </button>
```
Elena-brv, line 30: better make it a link

src/App.tsx
```diff
@@ -1,18 +1,46 @@
...
+    selectedTab: tabs[0],
+  };
+
+  chooseTab = (id: string) => {
+    const chosenTab = tabs.find(tab => tab.id === id) || tabs[0];
```
Elena-brv, line 23: you can use here findIndex method. Then no need to do 'find' again in a Tabs component, just use index. Or you can pass to Tabs not selectedTab.id but all selectedTab

Elena-brv (changes requested):
Content 🔥

Elena-brv (approved):

PR #543, opened 2021-09-12
React tabs

SeverusOwl:
[DEMO LINK](https://SeverusOwl.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 
 import './App.scss';
+import { TabsList } from './components/TabsList';
 
 const tabs: Tab[] = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,18 +1,44 @@
...
-    <div>{tabs.length}</div>
-  </div>
-);
+class App extends React.Component<{}, State> {
+  state = {
+    tabsCopy: [...tabs],
+    selectedTab: 'tab-1',
+  };
+
+  changeSelect = (tab: any) => {
```
nastyyvolk, line 18: I think type Tab should be used here instead of any

nastyyvolk (approved):
Well done 🔥

src/App.tsx
Elena-brv: don't use any

src/App.tsx
```diff
@@ -1,18 +1,44 @@
...
-const App: React.FC = () => (
-  <div className="App">
-    <h1>Selected tab is&nbsp;Tab 1</h1>
-    <div>{tabs.length}</div>
-  </div>
-);
+class App extends React.Component<{}, State> {
+  state = {
+    tabsCopy: [...tabs],
+    selectedTab: 'tab-1',
```
Elena-brv, line 15: if it's id, name it selectedTabId, and better here to store not id, but all tab. So you will be able not to do 'find' method many times

Elena-brv (changes requested):

src/components/TabsList/TabsList.tsx
```diff
@@ -0,0 +1,32 @@
...
+export const TabsList: React.FC<Props> = ({ tabs, selectedTab, onChange }) => {
+  return (
+    <>
+      <div className="buttons">
+        {tabs.map(tab => (
+          <button
+            className={classNames('button', selectedTab && 'selected')}
+            type="button"
+            name={tab.id}
+            onClick={() => onChange(tab.id)}
```
Elena-brv, line 20: add check 
image

src/App.tsx
```diff
@@ -1,18 +1,49 @@
...
+    return (
+      <div className="App">
+        <h1>{`Selected tab is ${selectedTab.title}`}</h1>
+        <div>
+          <TabsList
+            tabs={tabs}
+            selectedTab={selectedTab}
+            onChange={tabId => {
+              this.changeSelect(tabId);
+            }}
```
Elena-brv, line 41:
```suggestion
            onChange={this.changeSelect}
```

Elena-brv (changes requested):

Elena-brv (commented):
and add some styles for an active tab

Elena-brv (changes requested):
look to these comments 
image

Elena-brv (approved):

PR #555, opened 2021-09-14
React - Lifting state up - Tabs

george-w-boo:
[DEMO LINK](https://george-w-boo.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,9 @@
 import React from 'react';
 
+import 'bootstrap/dist/css/bootstrap.min.css';
 import './App.scss';
+
+import { Tabs } from './components/Tabs/Tabs';
 
 const tabs: Tab[] = [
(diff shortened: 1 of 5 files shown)


DmytroMykoliv (approved):
Good work _))

src/App.tsx
```diff
@@ -1,18 +1,50 @@
...
+type Props = {};
+
+type State = {
+  tabsFromServer: Tab[];
+  selectedTab: Tab;
+};
+
+class App extends React.Component<Props, State> {
+  state: State = {
+    tabsFromServer: tabs,
```
danheim, line 23: avoid `fromServer`, it doesn't matter here
```suggestion
    tabs: tabs,
```

danheim (approved):

src/App.tsx
george-w-boo: > avoid `fromServer`, it doesn't matter here

Great point, thanks ;)

PR #560, opened 2021-09-14
React Tabs Solution

ihor-shapoval:
[DEMO LINK](https://ihor-shapoval.github.io/react_tabs/)


Andrii-Pinchuk (approved):
Good job 👍 
Also, you can add utility "classnames" and give class "active" by click on tab. You can see how it works in my job.

danheim (approved):

PR #570, opened 2021-09-15
solution

derelicttt:
[DEMO LINK](https://derelicttt.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
 import React from 'react';
 
 import './App.scss';
+import { Tabs } from './components/Tabs';
 
 const tabs: Tab[] = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,18 +1,45 @@
...
+    });
+  };
+
+  render() {
+    const { selectedTab } = this.state;
+
+    return (
+      <div className="App">
+        <h1>{`Selected tab is ${selectedTab.title}`}</h1>
+        <div>{tabs.length}</div>
```
ivanchenko-kirill, line 35:
```suggestion

```
I don`t think you need this here

ivanchenko-kirill (approved):
You can add Bootstrap to make your page look more nice

src/components/Tabs.tsx
```diff
@@ -0,0 +1,28 @@
...
+export const Tabs: React.FC<Props> = ({ tabs, selectedTab, onTabSelected }) => (
+  <div className="container">
+    {tabs.map(({ title, id }) => (
+      <button
+        key={id}
+        type="button"
+        className={classNames({
+          active: id === selectedTab.id,
+        })}
+        onClick={() => onTabSelected(id)}
```
Elena-brv, line 19: image

derelicttt: I do not understand how to implement it

Elena-brv: you should check if current tab id is not the same as id of tab that was passed on props

Elena-brv (changes requested):

Elena-brv (approved):

PR #667, opened 2021-12-15
Solution

Vitaliy-0:
[demo](https://vitaliy-0.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,18 +1,42 @@
 import React from 'react';
+import { tabs } from './api/data';
 
 import './App.scss';
+import { Tabs } from './components/Tabs';
 
-const tabs: Tab[] = [
-  { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
-  { id: 'tab-2', title: 'Tab 2', content: 'Some text 2' },
-  { id: 'tab-3', title: 'Tab 3', content: 'Some text 3' },
-];
+type State = {
+  selectedTab: string,
+};
 
-const App: React.FC = () => (
-  <div className="App">
-    <h1>Selected tab is&nbsp;Tab 1</h1>
-    <div>{tabs.length}</div>
-  </div>
-);
+export class App extends React.Component<{}, State> {
+  state = {
+    selectedTab: 'tab-1',
+  };
 
-export default App;
+  onTabSelected = (selectedTab: Tab) => {
+    this.setState({
+      selectedTab: tabs.find(
+        tab => tab.id === selectedTab.id,
+      )?.id || 'tab-1',
+    });
+  };
+
+  render() {
+    const { selectedTab } = this.state;
+
+    return (
+      <div className="App">
+        <h1>{`Selected tab is ${tabs.find(tab => selectedTab === tab.id)?.title}`}</h1>
+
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -1,18 +1,42 @@
...
+      <div className="App">
+        <h1>{`Selected tab is ${tabs.find(tab => selectedTab === tab.id)?.title}`}</h1>
+
+        <Tabs
+          tabs={tabs}
+          onTabSelected={this.onTabSelected}
+        />
+
+        <p>
+          {tabs.find(tab => selectedTab === tab.id)?.content}
```
dariia-romanova, line 37: You can make variable for `tabs.find(tab => selectedTab === tab.id)` so you don't have to do it twice

src/components/Tabs.tsx
```diff
@@ -0,0 +1,25 @@
...
+  onTabSelected: (selectedTab: Tab) => void,
+};
+
+export const Tabs: React.FC<Props> = ({ tabs, onTabSelected }) => (
+  <div className="Tabs">
+    <nav className="Tabs__nav">
+      {tabs.map(tab => (
+        <div className="Tab" key={tab.id}>
+          <Tab
+            key={tab.id}
```
dariia-romanova, line 17: I think this key is unnecessary

dariia-romanova (approved):
Good job)

src/components/Tabs.tsx
Vitaliy-0: No, will be an error in console, need unique key property

src/App.tsx
```diff
@@ -1,18 +1,43 @@
...
+  };
 
-export default App;
+  onTabSelected = (selectedTab: Tab) => {
+    this.setState({
+      selectedTab: tabs.find(
+        tab => tab.id === selectedTab.id,
+      ) || tabs[0],
+    });
+  };
```
AlinaMatuschak, line 22: Can you use selectedTab like the actual tab?

```suggestion
  onTabSelected = (selectedTab: Tab) => {
    this.setState({
      selectedTab: selectedTab || tabs[0],
    });
  };
```

src/components/Tabs.tsx
```diff
@@ -0,0 +1,33 @@
...
+  <div className="Tabs">
+    <nav className="Tabs__nav">
+      {tabs.map(tab => (
+        <div className="Tab" key={tab.id}>
+          <a
+            className={classNames('Tab__link',
+              { 'Tab__link--selected': selectedTab.id === tab.id })}
+            href="/"
+            onClick={(event) => {
+              event.preventDefault();
```
AlinaMatuschak, line 22: Maybe better use the button? You won’t need to use preventDefault and add href attribute

src/components/Tabs.tsx
```diff
@@ -0,0 +1,33 @@
...
+
+              onTabSelected(tab);
```
AlinaMatuschak, line 24: Don't call the callback if the tab was not changed

AlinaMatuschak (changes requested):

AlinaMatuschak (approved):

PR #717, opened 2022-01-26
solution

Basil-Panasiuk:
- [DEMO LINK](https://Basil-Panasiuk.github.io/react_tabs/)


src/App.tsx
```diff
@@ -1,18 +1,43 @@
...
+type State = {
+  selectedTab: Tab;
+};
+
+class App extends React.Component<{}, State> {
+  state = {
+    selectedTab: tabs[0],
+  };
+
+  changeTab = (tab: Tab) => {
```
Olexandr6, line 21:
```suggestion
  handleTabChange = (tab: Tab) => {
```

src/components/Tabs.tsx
```diff
@@ -0,0 +1,46 @@
...
+  onTabSelected: (tab: Tab) => void;
+};
+
+class Tabs extends React.Component<Props> {
+  handleTabClick = (tab: Tab) => {
+    const { selectedTabId, onTabSelected } = this.props;
+
+    if (selectedTabId !== tab.id) {
+      onTabSelected(tab);
+    }
```
Olexandr6, line 15: You can do this check in the `handleTabChange` function, so move this logic there and delete this unnecessary `handleTabClick`

Olexandr6 (changes requested):
Good job 👍  Well done 🔥 But one small thing should be fixed, check comments👇

src/components/Tabs.tsx
Basil-Panasiuk: image

i did this because the rules forbid calling a callback if the tab was not changed

Denchepornyuk (approved):

PR #741, opened 2022-01-30
Solution

AlexAzalor:


Sesile26 (approved):

src/App.tsx
```diff
@@ -1,18 +1,48 @@
...
+  };
+
+  render() {
+    const { selectedTab } = this.state;
+
+    return (
+      <div className="App">
+        <h1 className="App__title">
+          Selected tab is
+          {' '}
```
Denchepornyuk, line 37: if you want to make some space between the text, it is better to do it with css

src/Tabs.tsx
```diff
@@ -0,0 +1,33 @@
...
+  handleSelector: (tab: string) => void;
+};
+
+export const Tabs: React.FC<Props> = ({ tabs, selectedTabId, handleSelector }) => {
+  const findTabContent = () => tabs.find(tab => tab.id === selectedTabId) || tabs[0];
+
+  return (
+    <div className="Tabs__tab">
+      {tabs.map(tab => (
+        <>
```
Denchepornyuk, line 16: do you need React.Fragment here?

src/Tabs.tsx
```diff
@@ -0,0 +1,33 @@
...
+            className={classNames('Tabs__button',
+              { 'Tabs__button--active': tab.id === selectedTabId })}
+            onClick={() => handleSelector(tab.id)}
+          >
+            {tab.title}
+          </button>
+        </>
+      ))}
+      <p className="Tabs__text">
+        {findTabContent().content}
```
Denchepornyuk, line 29: try not to write code like this
Here should be only content without function `findTabContent`

Denchepornyuk (changes requested):

Denchepornyuk (approved):

PR #751, opened 2022-02-01
Solution

Rmu4h:
[DEMO LINK](https://Rmu4h.github.io/react_tabs/)


src/components/Tabs.tsx
```diff
@@ -0,0 +1,43 @@
...
+            )
+            : (
+              <button
+                type="button"
+                onClick={() => onTabSelected(tab)}
+                className="nav-link"
+              >
+                {tab.title}
+              </button>
+            )}
```
Denchepornyuk, line 32: buttons are the same, add class `active` with condition rendering and remove condition rendering for buttons

Denchepornyuk (changes requested):

src/components/Tabs.tsx
```diff
@@ -11,25 +11,13 @@ export const Tabs: React.FC<Props> = ({ tabs, selectedTabId, onTabSelected }) =>
...
-                onClick={() => onTabSelected(tab)}
-                className="nav-link"
-              >
-                {tab.title}
-              </button>
-            )}
+          <button
+            type="button"
+            onClick={() => onTabSelected(tab)}
+            className={selectedTabId === tab.id ? 'nav-link active' : 'nav-link'}
```
Denchepornyuk, line 17: the best solution will be with classnames library

Denchepornyuk (approved):

PR #774, opened 2022-03-15
Solution

Vonatlus:
[DEMO](http://vonatlus.github.io/react_tabs/)


Elena-brv (changes requested):
image

cypress/integration/page.spec.js
```diff
@@ -1,4 +1,4 @@
-// WRITE TESTS HERE
+WRITE TESTS HERE
```
Elena-brv, line 1: it should be commented

Elena-brv (approved):

PR #786, opened 2022-04-03
Super difficult

DimaBrushnivskyi:


src/Tabs.tsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+
+const Tabs: React.FC<Tabs> = ({
```
Bohdan-Maksymiuk, line 3:
```suggestion
const Tabs: React.FC<Props> = ({
```

Bohdan-Maksymiuk: and define Props type above with callback and tabs

Bohdan-Maksymiuk (changes requested):

src/Tabs.tsx
```diff
@@ -0,0 +1,26 @@
...
+  tabs,
+  onTabSelected,
+}) => {
+  return (
+    <div className="Container">
+      {tabs.map((tab, index) => (
+        <div className="Tab">
```
Bohdan-Maksymiuk, line 10: you should use key when mapping

aliceblunt8 (approved):

PR #787, opened 2022-04-03
Selecting tabs

aspurgilus:
[DEMO LINK](https://aspurgilus.github.io/react_tabs/)

### DIFF
--- a/src/App.scss
+++ b/src/App.scss
@@ -1,1 +1,5 @@
-// Put styles here
+.App {
+  width: 600px;
+  margin: 0 auto;
+  padding-top: 60px;
+}
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,13 @@
-import React from 'react';
+import React, { useState } from 'react';
 
 import './App.scss';
+import { Tabs } from './Tabs';
+
+interface Tab {
+  id: string,
+  title: string,
+  content: string
+}
 
 const tabs: Tab[] = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
@@ -8,11 +15,23 @@ const tabs: Tab[] = [
   { id: 'tab-3', title: 'Tab 3', content: 'Some text 3' },
 ];
 
-const App: React.FC = () => (
-  <div className="App">
-    <h1>Selected tab is&nbsp;Tab 1</h1>
-    <div>{tabs.length}</div>
-  </div>
-);
+const App: React.FC = () => {
+  const [selectedTab, setSelected] = useState(tabs[0]);
+
+  const onTabSelected = (tab: Tab) => {
+    setSelected(tab);
+  };
+
+  return (
+    <div className="App">
+      <h1>{`Selected tab is ${selectedTab.title}`}</h1>
+      <Tabs
+        tabs={tabs}
+        selectedTabId={selectedTab.id}
+        onSelect={onTabSelected}
+      />
+    </div>
+  );
+};
 
 export default App;
--- a/src/Tabs.tsx
+++ b/src/Tabs.tsx
@@ -1,0 +1,45 @@
+import React from 'react';
+import classNames from 'classnames';
+
+interface Props {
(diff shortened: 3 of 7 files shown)


src/App.scss
```diff
@@ -1 +1,5 @@
-// Put styles here
+.App {
+  width: 600px;
+  margin: 0 auto;
```
jstmpelowycz, line 3: I wouldn't suggest positioning a separate block (BEM)

src/App.tsx
```diff
@@ -1,18 +1,40 @@
...
   { id: 'tab-3', title: 'Tab 3', content: 'Some text 3' },
 ];
 
-const App: React.FC = () => (
-  <div className="App">
-    <h1>Selected tab is&nbsp;Tab 1</h1>
-    <div>{tabs.length}</div>
-  </div>
-);
+const App: React.FC = () => {
```
jstmpelowycz, line 18: I recommend using `React.memo()` by default

src/App.tsx
```diff
@@ -1,18 +1,40 @@
...
+  const [selectedTab, setSelected] = useState(tabs[0]);
+
+  const onTabSelected = useCallback((tab: Tab) => {
+    setSelected(tab);
+  }, []);
```
jstmpelowycz, line 23:
```suggestion
  const onTabSelected = useCallback(setSelected, []);
```

src/Tabs.tsx
```diff
@@ -0,0 +1,32 @@
...
+      <div className="tabs">
+        <ul>
+          {tabs.map(tab => (
+            <li
+              key={tab.id}
+            >
+              <button
+                className={classNames('button', { 'is-info': selectedTabId === tab.id })}
+                type="button"
+                onClick={() => onSelect(tab)}
```
jstmpelowycz, line 22: You could have put the entire callback in a separate one applying `useCallback` so that re-rendering doesn't cause create a new handler

jstmpelowycz (approved):
A few moment to reflect upon, good job though😏

src/Tabs.tsx
aspurgilus: I didn't undestand this comment. Can you suggest please? What is callback might be look like?

aspurgilus:
> A few comments to reflect upon, good job though😏

Thanks. I implemented your suggestions except one that I replied.

src/Tabs.tsx
jstmpelowycz:
```
const handleButtonClick = useCallback(() => {
      if (!isSelected) {
        onTabSelected(tab);
      }
    }, [selectedTabId]);

    return (
      <>
        <button
          className="Tab__title"
          type="button"
          disabled={isSelected}
          onClick={handleButtonClick}
```

jstmpelowycz: All that is a just a matter of optimization. However, here is a recipe - instead of creating a callback inside `onClick` explicitly, you create a handler using `useCallback`

jstmpelowycz: NB: If you see, `isSelected` doesn't matter here, that's for UI

src/App.tsx
```diff
@@ -1,18 +1,38 @@
...
-const App: React.FC = () => (
-  <div className="App">
-    <h1>Selected tab is&nbsp;Tab 1</h1>
-    <div>{tabs.length}</div>
-  </div>
-);
+const App: React.FC = memo(() => {
+  const [selectedTab, setSelected] = useState(tabs[0]);
+
+  const onTabSelected = useCallback(setSelected, []);
```
Ihor-Karpyn, line 21: redundant

src/Tabs.tsx
```diff
@@ -0,0 +1,37 @@
+import React, { memo } from 'react';
+import classNames from 'classnames';
```
Ihor-Karpyn, line 2: Can short
```suggestion
import cn from 'classnames';
```

Ihor-Karpyn (approved):

PR #797, opened 2022-04-06
solution

AnnSerdechna:
[Demo](https://github.com/AnnSerdechna/react_movies-list-add-form.git)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
-import React from 'react';
-
+import { useCallback, useState } from 'react';
+import { Tabs } from './Tabs';
 import './App.scss';
 
 const tabs: Tab[] = [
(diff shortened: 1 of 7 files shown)


krolich500grn (approved):
Good job!

src/App.tsx
```diff
@@ -8,11 +8,26 @@ const tabs: Tab[] = [
...
 ];
 
-const App: React.FC = () => (
-  <div className="App">
-    <h1>Selected tab is&nbsp;Tab 1</h1>
-    <div>{tabs.length}</div>
-  </div>
-);
+const App: React.FC = () => {
+  const [selectedTab, setSelectedTab] = useState(tabs[0]);
```
krolich500grn, line 12: it must be id

```suggestion
  const [selectedTabId, setSelectedTabId] = useState(tabs[0].id);
```

krolich500grn (changes requested):

src/App.tsx
Ihor-Karpyn, line 12: better to save tab's id here, because you save tab in 2 places. In tabs array and here

Ihor-Karpyn (changes requested):

src/App.tsx
AnnSerdechna: This is condition of the task:
2 - Save the selectedTab in the App (the first one by default)
4 - Pass the selectedTabId as a prop to the Tabs, the specified tab should be selected if possible (otherwise the first tab is selected)

src/Tabs.tsx
```diff
@@ -0,0 +1,30 @@
+import { FC } from 'react';
+import classNames from 'classnames';
+
+type Props = {
+  tabs: Tab[];
+  selectedTabId: string;
+  onTabSelected: (tab: Tab) => void;
+};
+
+export const Tabs: FC<Props> = ({
```
solaryasha, line 10: use React.memo

src/Tabs.tsx
```diff
@@ -0,0 +1,30 @@
...
+}) => {
+  return (
+    <div className="Tab">
+      {tabs.map((tab) => (
+        <button
+          key={tab.id}
+          type="button"
+          onClick={() => onTabSelected(tab)}
+          className={classNames('Tab-btn',
+            { active: selectedTabId === tab.id })}
```
solaryasha, line 23: 🔥

solaryasha (approved):

PR #873, opened 2022-06-09
react tabs

Andriyk0:
[DEMO LINK](https://Andriyk0.github.io/react_tabs/)


SavMary (changes requested):
Your solution is good, but you didn't comply with the following conditions: 
Pass the selectedTabId as a prop to the Tabs,
Pass the onTabSelected callback to the Tabs.

Denchepornyuk (approved):

PR #879, opened 2022-06-10
Develop

KostiaO:
https://KostiaO.github.io/react_tabs/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,5 @@
-import React from 'react';
-
+import React, { useState } from 'react';
+import { Tabs } from './components/Tabs';
 import './App.scss';
 
 export const tabs: Tab[] = [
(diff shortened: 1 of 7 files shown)


KostiaO:
https://KostiaO.github.io/react_tabs/

KostiaO:
disabled 2 lines for eslint, cause using bulma, if needed, can improve

yelnikov-andrii (approved):

src/App.tsx
```diff
@@ -8,11 +8,30 @@ export const tabs: Tab[] = [
...
+    setSelectedTabId(id);
+  };
+
+  const tabList = [...tabs];
+
+  return (
+    <div className="App">
+      <h1>
+        Selected tab is&nbsp;
+        {tabList.find(tab => tab.id === selectedTabId)?.title}
```
Anna-Tkachenko, line 24: Will be better to store the result of find method in some variable and use this variable in jsx

src/components/Tabs.tsx
```diff
@@ -0,0 +1,46 @@
...
+interface Props {
+  tabList: Tab[],
+  selectedTabId: string,
+  onTabSelected: (id: string) => void
+}
+
+export const Tabs: React.FC<Props> = React.memo(
+  ({ tabList, onTabSelected, selectedTabId }) => {
+    const findedTab = useMemo(
+      () => tabList.find(tab => tab.id === selectedTabId),
```
Anna-Tkachenko, line 13: You call this method two times in two different components, will be better to call it once
You can store in App state full selected tab and pass only it's id to this component
It should help to avoid calling find method in both of components

Anna-Tkachenko (approved):

PR #882, opened 2022-06-10
solution

lytvyshko:
- [DEMO LINK](https://lytvyshko.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -10,17 +10,10 @@ export const tabs: Tab[] = [
 ];
 
 const App: React.FC = () => {
-  const [selectedTabId, setSelectedTabId] = useState('tab-1');
   const [selectedTab, setSelectedTab] = useState(tabs[0]);
 
-  const idSelect = (tab: Tab): void => {
-    setSelectedTabId(tab.id);
-  };
-
-  const findTabById = (id: string): void => {
-    const currentTab = tabs.find(tab => tab.id === id) || tabs[0];
-
-    setSelectedTab(currentTab);
+  const onTabSelected = (tab: Tab): void => {
+    setSelectedTab(tab);
   };
 
   return (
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -8,11 +9,32 @@ export const tabs: Tab[] = [
...
 
-const App: React.FC = () => (
-  <div className="App">
-    <h1>Selected tab is&nbsp;Tab 1</h1>
-    <div>{tabs.length}</div>
-  </div>
-);
+const App: React.FC = () => {
+  const [selectedTabId, setSelectedTabId] = useState('tab-1');
+  const [selectedTab, setSelectedTab] = useState(tabs[0]);
```
Anna-Tkachenko, line 14: You should save only selectedTab in state, selectedTabId is redundant, you can get needed value from selected tab

src/App.tsx
```diff
@@ -8,11 +9,32 @@ export const tabs: Tab[] = [
...
+    setSelectedTab(currentTab);
+  };
+
+  return (
+    <div className="App">
+      <h1>{`Selected tab is ${selectedTab.title}`}</h1>
+      <Tabs
+        tabs={tabs}
+        selectedTabId={selectedTabId}
+        selectedTab={selectedTab}
```
Anna-Tkachenko, line 32: You should pass only selectedTabId to the Tabs, selectedTab is redundant

Anna-Tkachenko (changes requested):

src/Tabs.tsx
```diff
@@ -1 +1,39 @@
...
+  const currentTab = findTabById(selectedTabId) || tabs[0];
+
+  return (
+    <>
+      <div className="tabs">
+        {tabs.map(tab => (
+          <button
+            key={tab.id}
+            type="button"
+            onClick={() => onTabSelected(tab)}
```
Anna-Tkachenko, line 28: Will be better to change selected tab only if selected tab id is not equal to the current tab id

Anna-Tkachenko (approved):

PR #883, opened 2022-06-11
solution

alexandra-protyanova:


yelnikov-andrii (approved):

src/components/Tabs.tsx
```diff
@@ -0,0 +1,39 @@
...
+      {tabs.map(tab => (
+        <div
+          className="container"
+          key={tab.id}
+        >
+          <button
+            className="button"
+            type="button"
+            onClick={() => {
+              onTabSelected(tab);
```
Anna-Tkachenko, line 27: Will be good to call this function only if tab id is not equal to selected tab id

Anna-Tkachenko (approved):

PR #884, opened 2022-06-11
SolutionI

EnjSadman:
[DEMO LINK](https://EnjSadman.github.io/react_tabs/)

### DIFF
--- a/src/Tabs.tsx
+++ b/src/Tabs.tsx
@@ -1,5 +1,5 @@
-// implement a component here
 import React from 'react';
+import './Tabs.scss';
 
 type Props = {
   tabsList : Tab[];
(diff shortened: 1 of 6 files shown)


src/Tabs.tsx
```diff
@@ -1 +1,41 @@
...
+          className="tab"
+        >
+          <button
+            type="button"
+            className="tab__pick"
+            onClick={() => {
+              onTabSelected(singleTab);
+            }}
+          >
+            {`Tab ${index + 1}`}
```
iwanttosmile4u, line 29: You can write here just singleTab.title, i guess. So you don't need index as parametr in map method.

src/Tabs.tsx
```diff
@@ -1 +1,41 @@
...
+          </button>
+          {selectedTabId === singleTab.id
```
iwanttosmile4u, line 31: I think, that tab content should be outside the div with button. So the logic, witch responsible for showing tab text, should be differnt.

src/Tabs.tsx
```diff
@@ -1 +1,40 @@
...
+        <>
+          <div
+            key={singleTab.id}
+            className="tab"
+          >
+            <button
+              type="button"
+              className="tab__pick"
+              onClick={() => {
+                onTabSelected(singleTab);
```
Anna-Tkachenko, line 27: Will be better to call this function only if singleTab id is not equal to selected tab id

Anna-Tkachenko (approved):

PR #888, opened 2022-06-12
tabs on react

kitaminary:
[DEMO LINK](https://<your_account>.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
-
 import './App.scss';
+import { useState } from 'react';
+import { Tab } from './react-app-env';
+import { Tabs } from './components/Tabs/Tabs';
 
 export const tabs: Tab[] = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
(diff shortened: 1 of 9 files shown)


src/App.tsx
```diff
@@ -1,18 +1,39 @@
...
+    setContent(tabContent[0].content);
+  };
+
+  const selectedTabId = (tabId: string) => (
+    setTitleId(tabId)
+  );
+
+  return (
+    <div className="tab">
+      <h1 className="tab__title title m-6">{`Selected tab is ${(tabs.filter(tab => tab.id === titleId))[0].title}`}</h1>
```
Anna-Tkachenko, line 28: Will be better to store selected tab in the state and use that variable in jsx

src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,49 @@
...
+export const Tabs: React.FC<Props> = ({
+  tabs,
+  selectedTabId,
+  content,
+  onTabSelected,
+}) => (
+  <div className="tab__box m-6">
+    <div className="tab__buttons-box is-flex-direction-column">
+      {tabs.map((tab: Tab) => {
+        return (
```
Anna-Tkachenko, line 22: return is redundant

src/components/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,49 @@
...
+              },
+            )}
+          >
+            <button
+              type="button"
+              className="has-text-info button is-white"
+              onClick={() => {
+                selectedTabId(tab.id);
+                onTabSelected(tab.id);
+              }}
```
Anna-Tkachenko, line 39: Will be better to call this methods only if selected tab id is not equal to current tab id

Anna-Tkachenko (changes requested):

Denchepornyuk (approved):

PR #890, opened 2022-06-13
made tabs select

uliton:
[DEMO LINK](https://uliton.github.io/react_tabs/)


src/App.tsx
```diff
@@ -1,18 +1,25 @@
...
-    <div>{tabs.length}</div>
-  </div>
-);
+const App: React.FC = () => {
+  const tabsCopy = [...tabs];
+
+  return (
+    <div className="App">
+      <Tabs
+        tabs={tabsCopy}
```
Anna-Tkachenko, line 19: You should store selectedTabId in the App component, and pass this value as props to the Tabs component

Denchepornyuk (approved):

PR #897, opened 2022-06-14
Selected tabs

Oleksandr-Ivanchenko:
[DEMO LINK](https://Oleksandr-Ivanchenko.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,6 @@
-import React from 'react';
-
+import React, { useState } from 'react';
 import './App.scss';
+import { Tabs } from './Tabs';
 
 export const tabs: Tab[] = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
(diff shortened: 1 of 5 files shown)


src/Tabs.tsx
```diff
@@ -1 +1,22 @@
...
+  onTabSelect: (tabId: string) => void,
+}
+
+export const Tabs: React.FC<Props> = ({ tabs, onTabSelect }) => {
+  return (
+    <>
+      {tabs.map((tab: Tab) => (
+        <button
+          type="button"
+          onClick={() => onTabSelect(tab.id)}
```
Anna-Tkachenko, line 15: Will be better to call this function only if tab id is now equal to selected tab id

src/App.tsx
```diff
@@ -1,18 +1,39 @@
...
 
-const App: React.FC = () => (
-  <div className="App">
-    <h1>Selected tab is&nbsp;Tab 1</h1>
-    <div>{tabs.length}</div>
-  </div>
-);
+const App: React.FC = () => {
+  const [selectedTabId, setSelectedTabId] = useState(tabs[0].id);
+  const [selectedTab, setSelectedTab] = useState(tabs[0]);
```
Anna-Tkachenko, line 13: Let's store only id and will find selected tab by this id

src/App.tsx
```diff
@@ -1,18 +1,39 @@
...
+
+  const onTabSelect = (tabId: string) => {
+    const currentSelectedTab = tabs.find(tab => tab.id === tabId);
+
+    setSelectedTabId(tabId);
+    setSelectedTab(currentSelectedTab || tabs[0]);
```
Anna-Tkachenko, line 19: it's redundant

Anna-Tkachenko (changes requested):

src/Tabs.tsx
```diff
@@ -1 +1,27 @@
...
+  return (
+    <>
+      {tabs.map((tab: Tab) => (
+        <button
+          type="button"
+          onClick={() => {
+            if (selectedTabId !== tab.id) {
+              onTabSelect(tab.id);
+            }
+          }}
```
Denchepornyuk, line 20: create a separate function for this onClick

Denchepornyuk (approved):

PR #898, opened 2022-06-14
Solution

AndriySimyachenko:
[DEMO LINK](https://<your_account>.github.io/react_tabs/)

### DIFF
--- a/src/Tabs.tsx
+++ b/src/Tabs.tsx
@@ -1,1 +1,38 @@
-// implement a component here
+import React from 'react';
+import { Tab } from './react-app-env';
+
+type Props = {
+  tabs: Tab[],
+  selectedTabId: string,
+  onTabSelected: (tab: Tab) => void
+};
+
+export const Tabs: React.FC<Props> = ({
+  tabs,
+  selectedTabId,
+  onTabSelected,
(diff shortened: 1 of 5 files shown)


alexandra-protyanova (approved):

src/Tabs.tsx
```diff
@@ -1 +1,38 @@
...
+}) => {
+  return (
+    <div>
+      {tabs.map(tab => (
+        <div key={tab.id}>
+
+          <button
+            type="button"
+            onClick={() => {
+              onTabSelected(tab);
```
Anna-Tkachenko, line 23: Will be better to change selected tab only if selected tab id is not equal to current tab id

src/Tabs.tsx
```diff
@@ -1 +1,38 @@
...
+            }}
+          >
+            {tab.title}
+          </button>
+
+        </div>
+      ))}
+
+      <div>
+        {tabs.find(tab => tab.id === selectedTabId)?.content}
```
Anna-Tkachenko, line 33: Will be better to store the result of the find method in some variable and use this variable in jsx

Anna-Tkachenko (approved):

PR #901, opened 2022-06-16
Develop

Danylo-Dziuban:
[DEMO LINK](https://Danylo-Dziuban.github.io/react_tabs/)

### DIFF
--- a/src/Tabs.tsx
+++ b/src/Tabs.tsx
@@ -1,1 +1,48 @@
-// implement a component here
+/* eslint-disable max-len */
+import { Button, Box } from '@mui/material';
+
+interface Ta {
+  id: string,
+  title: string,
+  content: string,
+}
+
+type Props = {
+  tabs: Ta[],
+  onSelect: (id: string) => void,
+  selectedTabId: string,
+};
+
+export const Tabs: React.FC<Props> = ({
+  tabs,
+  selectedTabId,
+  onSelect,
+}) => {
+  return (
+    <Box
+      data-cy="tab-content"
+    >
+      <Box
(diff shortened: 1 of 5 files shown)


src/Tabs.tsx
```diff
@@ -1 +1,48 @@
-// implement a component here
+/* eslint-disable max-len */
+import { Button, Box } from '@mui/material';
+
+interface Ta {
```
Anna-Tkachenko, line 4: What is Ta?

src/Tabs.tsx
```diff
@@ -1 +1,48 @@
...
+      <Box
+        component="div"
+        sx={{ display: 'flex', flexDirection: 'row' }}
+      >
+        {tabs.map((tab: Ta) => (
+          <Box>
+            <Button
+              type="button"
+              onClick={
+                () => (onSelect(tab.id))
```
Anna-Tkachenko, line 34: Will be better to change selected tab id only if selected tab id is not equal to the current tab

src/Tabs.tsx
```diff
@@ -1 +1,48 @@
...
+              {tab.title}
+            </Button>
+          </Box>
+        ))}
+      </Box>
+
+      <div>
+        {tabs.find((tab: Ta) => tab.id === selectedTabId)?.content || tabs[0].content}
+      </div>
+    </Box>
```
Anna-Tkachenko, line 46: You wrote this function in parent component
Will be better to pass needed props to the children component, if it's possible

Anna-Tkachenko (approved):

PR #906, opened 2022-06-22
solution

iLebo-stack:
[DEMO LINK](https://iLebo-stack.github.io/react_tabs/)

### DIFF
--- a/src/Tabs.tsx
+++ b/src/Tabs.tsx
@@ -1,1 +1,37 @@
-// implement a component here
+import React from 'react';
+import { Tab } from './react-app-env';
+
+interface Props {
+  tabs: Tab[];
+  selectedTabId: string;
+  onTabSelected: (tab: Tab) => void;
+}
+
+export const Tabs: React.FC<Props> = ({
+  tabs,
+  selectedTabId,
+  onTabSelected,
+}) => {
+  return (
+    <div>
+      {tabs.map(tab => (
+        <React.Fragment key={tab.id}>
+          <button
+            type="button"
+            onClick={() => {
+              if (selectedTabId === tab.id) {
+                return;
+              }
+
+              onTabSelected(tab);
+            }}
+          >
+            {tab.title}
+          </button>
+          {tab.id === selectedTabId
+          && <p data-cy="tab-content">{tab.content}</p>}
+        </React.Fragment>
+      ))}
+    </div>
+  );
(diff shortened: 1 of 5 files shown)


src/Tabs.tsx
```diff
@@ -1 +1,54 @@
...
+                href="#"
+                type="button"
+                onClick={() => {
+                  if (selectedTabId === tab.id) {
+                    return;
+                  }
+
+                  onTabSelected(tab);
+                }}
+              >
```
efdalyalcin, line 39: Nice implementation 👍

efdalyalcin (approved):
Well done

src/Tabs.tsx
```diff
@@ -1 +1,54 @@
...
+      <div className="tabs is-boxed">
+        <ul>
+          {tabs.map(tab => (
+            <li
+              key={tab.id}
+              className={
+                selectedTabId === tab.id
+                  ? 'is-active'
+                  : ''
+              }
```
meforoma, line 27: better do this
```suggestion
              className={ 'is-active': selectedTabId === tab.id}
```

src/Tabs.tsx
```diff
@@ -1 +1,54 @@
...
+              </a>
+            </li>
+          ))}
+        </ul>
+      </div>
+      {tabs.map(tab => (
+        <React.Fragment key={tab.id}>
+          {tab.id === selectedTabId
+          && <p data-cy="tab-content">{tab.content}</p>}
+        </React.Fragment>
```
meforoma, line 50: why `.Fragment`?
this should do just fine:
```suggestion
        {tab.id === selectedTabId && (
          <p data-cy="tab-content">{tab.content}</p>
        )}
```

meforoma (changes requested):
good one 👍 
please consider few comments

src/Tabs.tsx
```diff
@@ -1 +1,48 @@
...
+  selectedTabId: string;
+  onTabSelected: (tab: Tab) => void;
+}
+
+export const Tabs: React.FC<Props> = ({
+  tabs,
+  selectedTabId,
+  onTabSelected,
+}) => {
+  return (
```
ihor-jpeg, line 17: does not seem like you need a return keyword here

ihor-jpeg (approved):

PR #924, opened 2022-07-11
React tabs

vsupruniuk:
[DEMO LINK](https://vsupruniuk.github.io/react_tabs/)


Olexandr6 (changes requested):
Your solution is good👍 Let's improve it a bit

Olexandr6 (approved):

PR #947, opened 2022-07-20
Solution created

MariiaNH:
[DEMO LINK](https://mariianh.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,5 +1,7 @@
-import React from 'react';
+import { FC, useState } from 'react';
 import './App.scss';
+import { Tabs } from './components/Tabs/Tabs';
+import Tab from './types/Tab';
 
 export const tabs = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,38 +1,37 @@
...
+      setChoosedTab(tab);
+    }
+  };
+
   return (
     <div className="section">
       <h1 className="title">
-        Selected tab is&nbsp;Tab 1
+        Selected tab is&nbsp;
+        {choosedTab.title}
```
jstmpelowycz, line 25: here, you'd rather use interpolation – it is easier to implement + it will make it look simpler

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,34 @@
...
+  choosedTab: Tab
+}
+
+export const Tabs: FC<Props> = ({ tabs, chengeTab, choosedTab }) => (
+  <div className="tabs">
+    <ul className="tabs__list">
+      {tabs.map(tab => (
+        <li
+          key={tab.id}
+          className={classNames(tab.id === choosedTab.id ? 'is-active' : '')}
```
jstmpelowycz, line 17:
```suggestion
          className={classNames({
            'is-active' : tab.id === choosedTab.id
          })}
```

jstmpelowycz: here, you apply `classNames` but don't actually use it 🙂

jstmpelowycz (changes requested):

PR #990, opened 2022-08-01
solution

FANTAZER-NURE:
https://FANTAZER-NURE.github.io/react_tabs/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,6 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
+import { Tabs } from './components/Tabs/Tabs';
 
 export const tabs = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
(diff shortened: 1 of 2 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,62 @@
...
+interface Tab {
+  id: string,
+  title: string,
+  content: string,
+}
+
+type Props = {
+  tabs: Tab[],
+  selectedTabId: string,
+  handleClick: any,
```
valerii-derkach, line 12: do not spicify any. it is a function that returns void

valerii-derkach (changes requested):

src/App.tsx
```diff
@@ -9,34 +10,22 @@ export const tabs = [
...
-        Selected tab is&nbsp;Tab 1
-      </h1>
-
-      <div>
-        <div className="tabs is-boxed">
-          <ul>
-            <li className="is-active">
-              <a href="#tab-1">Tab 1</a>
-            </li>
+  const [selectedTabId, setTab] = useState(tabs[0].id);
```
Anna-Tkachenko, line 13:
```suggestion
  const [selectedTabId, setSelectedTabId] = useState(tabs[0].id);
```

src/App.tsx
```diff
@@ -9,34 +10,22 @@ export const tabs = [
...
 
-            <li className="">
-              <a href="#tab-2">Tab 2</a>
-            </li>
+  const onClickTabItem = (id: string) => {
+    const tab = tabs.find(el => el.id === id);
```
Anna-Tkachenko, line 16: Pay attention to variable names, what is el?

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,62 @@
...
+                className={classNames(
+                  { 'is-active': selectedTabId === tab.id },
+                )}
+                key={tab.id}
+              >
+                <a
+                  href={`#${tab.id}`}
+                  id={tab.id}
+                  onClick={(e) => {
+                    e.preventDefault();
```
Anna-Tkachenko, line 42: prevent default is redundant here, this event don't have specific actions on click (as submit event, for example)

src/App.tsx
Anna-Tkachenko: And will be better to change selected tab id only if new tab id is not equal to current tab id

Anna-Tkachenko (changes requested):

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,61 @@
...
+interface Tab {
+  id: string,
+  title: string,
+  content: string,
+}
+
+type Props = {
+  tabs: Tab[],
+  selectedTabId: string,
+  handleClick: (arg: string)=>void,
```
Bohdan-Maksymiuk, line 12:
```suggestion
  handleClick: (arg: string) => void,
```

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,61 @@
...
+};
+
+export const Tabs:React.FC<Props> = (
+  {
+    tabs,
+    selectedTabId,
+    handleClick,
+  },
+) => {
```
Bohdan-Maksymiuk, line 21:
```suggestion
export const Tabs:React.FC<Props> = ({
    tabs,
    selectedTabId,
    handleClick,
  }) => {
```

Bohdan-Maksymiuk (approved):

PR #1000, opened 2022-08-13
solution

oleksandr-kotliarov:
[DEMO LINK](https://oleksandr-kotliarov.github.io/react_tabs/)

### DIFF
--- a/src/Tab.ts
+++ b/src/Tab.ts
@@ -1,0 +1,5 @@
+export interface Tab {
+  id: string;
+  title: string;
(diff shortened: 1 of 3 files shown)


src/Tab.ts
```diff
@@ -0,0 +1,5 @@
+export interface Tab {
```
VETALEK, line 1: better move this file to some kind of `types` folder 😉

VETALEK (approved):
🚀

PR #1012, opened 2022-08-16
Solution

LeonidPustovit:
[DEMO LINK](https://LeonidPustovit.github.io/react_tabs/)

### DIFF
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -1,1 +1,67 @@
-export const Tabs = () => {};
+import classNames from 'classnames';
+
+interface Tab {
+  id: string;
+  title: string;
+  content: string;
+}
+
+interface Props {
+  tabs: Tab[];
+  selectedTabId: string;
+  onTabSelected: (tabId: string) => void;
+}
+
+export const Tabs: React.FC<Props> = ({
+  tabs,
+  selectedTabId,
+  onTabSelected,
+}) => {
+  const selectedTab = tabs.find(tab => tab.id === selectedTabId) || tabs[0];
+  const handleTabSelect = (tabId: string) => {
+    if (!tabs.find(tab => tab.id === tabId)) {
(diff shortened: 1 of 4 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,67 @@
...
+    if (!tabs.find(tab => tab.id === tabId)) {
+      onTabSelected(tabs[0].id);
+    }
+
+    if (tabId === selectedTabId) {
+      return;
+    }
+
+    onTabSelected(tabId);
+  };
```
solaryasha, line 31: Let's use this logic to onTabSelected function

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,67 @@
...
+          <a href="#tab-1" onClick={() => onTabSelected}>Tab 1</a>
+        </li>
+
+        <li className="">
+          <a href="#tab-2">Tab 2</a>
+        </li>
+
+        <li className="">
+          <a href="#tab-3">Tab 3</a>
+        </li> */}
```
solaryasha, line 59: Don't leave comments

solaryasha (changes requested):

src/components/Tabs/Tabs.tsx
LeonidPustovit: As I understand this, I moved logic to App.tsx, and send onTabSelected already with this logic. But, I can't move second if, as I need it to prevent onTabSelected from being reinvoked if current tab is active.

ivan-kadykalo (approved):
🚀

ivan-kadykalo (changes requested):
Text are in the wrong place
image
You should fix it

LeonidPustovit:
Guess this one was present from the very beginning -_-

ivan-kadykalo (approved):

PR #1022, opened 2022-08-16
solution

Yura-Lysenko:


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,51 @@
...
+      onTabSelected(tab);
+    }
+  };
+
+  return (
+    <div>
+      <div className="tabs is-boxed">
+        <ul>
+          {tabs.map(tab => {
+            return (
```
kata96451, line 27: you don't need return here :) you can get shorter code if you write 
```
tabs.map(tab => (
  jsx
)
```

kata96451 (approved):
good 👍

solaryasha (approved):

PR #1036, opened 2022-08-17
Develop

blcts:
[DEMO LINK](https://blcts.github.io/react_tabs/)

### DIFF
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -23,7 +23,7 @@ export const Tabs: React.FC<Props> = (props) => {
         <ul>
           {tabs.map(tab => (
             <li
-              className={cn(`${(tab.id === activeTabId) && 'is--active'}`)}
+              className={cn(`${(tab.id === activeTabId) && 'is-active'}`)}
               key={tab.id}
             >
               <a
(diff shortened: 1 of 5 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,49 @@
...
+              className={cn(`${(tab.id === activeTabId) && 'is-active'}`)}
+              key={tab.id}
+            >
+              <a
+                href={`#${tab.id}`}
+                onClick={() => {
+                  if (tab.id !== activeTabId) {
+                    onTabSelected(tab);
+                  }
+                }}
```
solaryasha, line 35: Let's move this logic to the App component

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,49 @@
...
+              >
+                {tab.title}
+              </a>
+            </li>
+          ))}
+        </ul>
+      </div>
+
+      <div className="block" data-cy="tab-content">
+        {selectedTab[0].content}
```
solaryasha, line 45: What if we get an empty array after filtering ?

solaryasha (changes requested):

solaryasha (approved):

PR #1053, opened 2022-09-12
solution

Githi54:
[DEMO LINK](https://githi54.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,42 +1,35 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
+import { Tabs } from './components/Tabs';
+import { Tab } from './types/Tab';
 
-export const tabs = [
+export const tabs: Tab[] = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
   { id: 'tab-2', title: 'Tab 2', content: 'Some text 2' },
   { id: 'tab-3', title: 'Tab 3', content: 'Some text 3' },
 ];
 
 export const App: React.FC = () => {
+  const firstTab = tabs[0];
+
+  const [selectedTabId, setSelectedTabId] = useState(firstTab.id);
+  const selectedTab = tabs.find(tab => tab.id === selectedTabId) || firstTab;
+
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,42 +1,35 @@
...
-          </ul>
-        </div>
-
-        <div className="block" data-cy="TabContent">
-          Some text 1
-        </div>
-      </div>
+      <Tabs
+        tabs={tabs}
+        selectedTab={selectedTab}
```
sergii-nosachenko, line 30:
```suggestion
        selectedTabId={selectedTabId}
```

\#4 of task condition says: Pass the **selectedTabId** as a prop to the Tabs, the specified tab should be selected if possible (otherwise the first tab is selected).

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,49 @@
...
+import { Tab } from '../../types/Tab';
+
+type Props = {
+  tabs: Tab[];
+  selectedTab: Tab;
+  onSelected: (tab: Tab) => void;
+};
+
+export const Tabs: React.FC<Props> = ({ tabs, selectedTab, onSelected }) => {
+  return (
```
sergii-nosachenko, line 10:
```suggestion
export const Tabs: React.FC<Props> = ({ tabs, selectedTabId, onSelected }) => {
  // Check here if selected tab is ok
  return (
```
You need to add additional check for selectedTab here (as you did in App component). Then your tests will be passed successfully.

sergii-nosachenko (changes requested):
Hi, Mykyta. You have missed one condition from the task. That's why some of your tests failed. Fix this problem, please.
Rest of the code is ok 👌

sergii-nosachenko (approved):
Good work 🔥

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,59 @@
...
+                <a
+                  href={`#${id}`}
+                  data-cy="TabLink"
+                  onClick={() => {
+                    if (selectedTab.id === tab.id) {
+                      return;
+                    }
+
+                    onTabSelected(tab);
+                  }}
```
dellachaise, line 44: Move this code to a separate function and use it here.

dellachaise (approved):
Well done 👍

PR #1056, opened 2022-09-12
Solution

Dmytro-Matsehora:
[DEMO LINK](https://Dmytro-Matsehora.github.io/react_tabs/)

### DIFF
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -1,1 +1,51 @@
-export const Tabs = () => {};
+import { Tab } from '../../Tab';
+
+type Props = {
+  tabs: Tab[];
+  selectedTabId: string;
+  onTabSelected: (tab: Tab) => void;
+};
+
+export const Tabs: React.FC<Props> = ({
+  tabs,
+  selectedTabId,
+  onTabSelected,
+}) => {
+  const selectedTab = tabs.find(tab => tab.id === selectedTabId) || tabs[0];
+
+  return (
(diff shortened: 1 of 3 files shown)


tiserett (approved):
🔥🔥🔥

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,51 @@
...
+      <div className="tabs is-boxed">
+        <ul>
+          {tabs.map(tab => {
+            return (
+              <li
+                className={
+                  selectedTab?.id === tab.id
+                    ? 'is-active'
+                    : ''
+                }
```
dellachaise, line 27: Better install [classnames](https://www.npmjs.com/package/classnames) and use it instead of ternary operator.
The ternary operator is very limited in its functionality to be used in classNames. (this way usually did in real projects and would be great to work with it now and would be needed to do in test tasks).

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,51 @@
...
+                data-cy="Tab"
+                key={tab.id}
+              >
+                <a
+                  href={`#${tab.id}`}
+                  data-cy="TabLink"
+                  onClick={() => (
+                    selectedTabId !== tab.id && onTabSelected(tab)
+                  )}
```
dellachaise, line 36: Create a separate function and use it here.

dellachaise (changes requested):
Great job 👍 

Just add changes according to comments 👇

PR #1068, opened 2022-09-19
solution 1

kostya-kosyuk:
[DEMO LINK](https://kostya-kosyuk.github.io/react_tabs/)

### DIFF
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -1,1 +1,48 @@
-export const Tabs = () => {};
+import classNames from 'classnames';
+import { Tab } from '../../types/Tab';
+
+type Props = {
+  tabs: Tab[];
+  selectedTabId: string;
+  handleSelect: (tab: Tab) => void;
+};
+export const Tabs: React.FC<Props> = ({
+  tabs,
+  selectedTabId,
+  handleSelect,
+}) => {
+  return (
+    <div data-cy="TabsComponent">
+      <div className="tabs is-boxed">
+        <ul>
+          {tabs.map(tab => {
+            const { id, title } = tab;
+
+            return (
+              <li
+                className={classNames(
+                  selectedTabId === id && 'is-active',
+                )}
+                data-cy="Tab"
+              >
+                <a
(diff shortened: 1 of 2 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,48 @@
...
+          {tabs.map(tab => {
+            const { id, title } = tab;
+
+            return (
+              <li
+                className={classNames(
+                  selectedTabId === id && 'is-active',
+                )}
+                data-cy="Tab"
+              >
```
NataliyaAn, line 27: You forgot to add "key"

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,48 @@
...
+            );
+          })}
+        </ul>
+      </div>
+      <div
+        className="block"
+        data-cy="TabContent"
+      >
+        {tabs.find(tab => selectedTabId === tab.id)?.content}
+      </div>
```
NataliyaAn, line 45: This block would be better move in App, because "selectedTab" is already there and you can simply write "selectedTab.content" . Now you are looking for "selectedTab" in App , and then you are looking for it in "Tabs"

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,48 @@
...
+  selectedTabId: string;
+  handleSelect: (tab: Tab) => void;
+};
+export const Tabs: React.FC<Props> = ({
+  tabs,
+  selectedTabId,
+  handleSelect,
+}) => {
+  return (
+    <div data-cy="TabsComponent">
```
NataliyaAn, line 15: you should use React Fragment when you have blocks on the same level. You added an additional div block to the layout

NataliyaAn (approved):
Good job)
With a few fixes it will be better

src/components/Tabs/Tabs.tsx
AlinaMatuschak: Screenshot 2022-09-26 at 23 04 29

AlinaMatuschak (approved):

PR #1076, opened 2022-09-22
react-tabs soluton

NataliyaAn:
[DEMO LINK](https://NataliyaAn.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,40 +1,40 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
-
-export const tabs = [
-  { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
-  { id: 'tab-2', title: 'Tab 2', content: 'Some text 2' },
-  { id: 'tab-3', title: 'Tab 3', content: 'Some text 3' },
-];
+import { Tabs } from './components/Tabs';
+import { Tab } from './type';
+import { tabs } from './tabsFromServer';
 
 export const App: React.FC = () => {
+  const tabsArr = [...tabs];
+  const [isActiveId, onTabSelected] = useState('tab-1');
+
+  const setActiveId = (tab:Tab) => {
+    if (isActiveId === tab.id) {
+      return;
+    }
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,40 +1,40 @@
...
-  { id: 'tab-2', title: 'Tab 2', content: 'Some text 2' },
-  { id: 'tab-3', title: 'Tab 3', content: 'Some text 3' },
-];
+import { Tabs } from './components/Tabs';
+import { Tab } from './type';
+import { tabs } from './tabsFromServer';
 
 export const App: React.FC = () => {
+  const tabsArr = [...tabs];
+  const [isActiveId, onTabSelected] = useState('tab-1');
```
anna-kolmakova, line 11: I think it will be better to use useState(tabs[0].id)

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,40 @@
...
+import classNames from 'classnames';
+import { Tab } from '../../type';
+
+type Props = {
+  tabs:Tab[]
+  isActiveId: string;
+  setActiveId:(tab: Tab) => void;
+
+};
+
```
anna-kolmakova, line 11: You need to add here (on tabs)  same const as on App: 

const activeTab = tabsArr.find(tab => tab.id === isActiveId) || tabsArr[0]; 

Because: 
Point 4: Pass the selectedTabId as a prop to the Tabs, the specified tab should be selected if possible (otherwise the first tab is selected).

anna-kolmakova (approved):
Good job! I think if you fix point 4, all tests will be passed successfully

src/App.tsx
```diff
@@ -1,40 +1,40 @@
...
+        <Tabs
+          tabs={tabsArr}
+          isActiveId={isActiveId}
+          setActiveId={setActiveId}
+        />
 
         <div className="block" data-cy="TabContent">
-          Some text 1
+          {activeTab.content}
         </div>
```
AlinaMatuschak, line 38: Screenshot 2022-09-27 at 12 56 52

AlinaMatuschak (changes requested):
fix @anna-kolmakova comments )

AlinaMatuschak (approved):

PR #1089, opened 2022-09-24
Solution

kir8mir:
- [DEMO LINK](https://kir8mir.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,8 @@
-import React from 'react';
+import { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
+import { Tabs } from './components/Tabs';
 
 export const tabs = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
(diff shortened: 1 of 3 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,55 @@
...
+                href={`#${tab.id}`}
+                onClick={() => {
+                  setTabId(tab.id);
+                }}
+              >
+                {tab.title}
+              </a>
+            </li>
+          ))}
+        </ul>
```
Kulchevych, line 47:
```suggestion
          {tabs.map(({ title, id }) => (
            <li
              className={classNames(
                { 'is-active': selectedTabId === id },
              )}
              data-cy="Tab"
            >
              <a
                data-cy="TabLink"
                href={`#${id}`}
                onClick={() => {
                  setTabId(id);
                }}
              >
                {title}
              </a>
            </li>
          ))}
        </ul>
```
You can use destructurization

Kulchevych (approved):
Gj

src/components/Tabs/Tabs.tsx
dellachaise: Agree 👍

src/App.tsx
```diff
@@ -10,33 +11,21 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [selectedTabId, setTabId] = useState('tab-1');
```
dellachaise, line 14: Better use `tabs[0].id` because this way it would work with different data. If id would change your solution would need some changes.
```suggestion
  const [selectedTabId, setTabId] = useState(tabs[0].id);
```

src/App.tsx
```diff
@@ -10,33 +11,21 @@ export const tabs = [
...
-            </li>
-          </ul>
-        </div>
-
-        <div className="block" data-cy="TabContent">
-          Some text 1
-        </div>
-      </div>
+      <Tabs
+        selectedTab={selectedTab}
```
dellachaise, line 24: You don't need to pass this prop, because you could get access to the selected tab, because you have `tabs` and `selectedTabId`.
This line is redundund.

src/App.tsx
```diff
@@ -10,33 +11,21 @@ export const tabs = [
...
+        setTabId={setTabId}
```
dellachaise, line 25: Better leave `setTabId` in App, and create a separate handler method  to change tab id and pass it here as props.

[more details in video](https://mate.academy/learn/react/lifting-state-up#/video)

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,55 @@
...
+  selectedTabId,
+  tabs,
+}) => {
+  return (
+    <div data-cy="TabsComponent">
+      <div className="tabs is-boxed">
+        <ul>
+          {tabs.map(tab => (
+            <li
+              className={classNames(
```
dellachaise, line 31: Key?

dellachaise (changes requested):
Good start 👍 

Just add changes according to comments 👇

src/App.tsx
```diff
@@ -10,33 +11,24 @@ export const tabs = [
...
-          </ul>
-        </div>
-
-        <div className="block" data-cy="TabContent">
-          Some text 1
-        </div>
-      </div>
+      <Tabs
+        setTabId={handleTabId}
+        selectedTabId={selectedTabId}
```
AlinaMatuschak, line 29: Can you use setTabId here?
```suggestion
        selectedTabId={setTabId}
```

AlinaMatuschak (approved):

PR #1102, opened 2022-09-26
solution

yevhenii-podplietnii:
[DEMO LINK](https://yevhenii-podplietnii.github.io/react_tabs/)

### DIFF
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -1,1 +1,44 @@
-export const Tabs = () => {};
+import React from 'react';
+import classNames from 'classnames';
+import { Tab } from '../../types/Tab';
+
+type Props = {
+  tabs: Tab[];
+  selectedTab: Tab;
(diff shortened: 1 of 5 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,44 @@
...
+            const isSelected = selectedTab.id === id;
+
+            return (
+              <li
+                key={id}
+                className={classNames({ 'is-active': isSelected })}
+                data-cy="Tab"
+              >
+                <a
+                  onClick={() => !isSelected && onTabSelected(tab)}
```
dellachaise, line 27: Create a separate method instead of creating an inline event handler.

But interesting solution 👍

dellachaise (changes requested):
Great job 👍 

Just create a separate method instead of creating an inline event handler.

dellachaise (approved):
Well done 👍

PR #1118, opened 2022-09-27
add: tabs

ShmelyovOleg:
[DEMO LINK](https://ShmelyovOleg.github.io/react_tabs/)


vladgusakovskiy (approved):
Great 👍

src/App.tsx
```diff
@@ -10,33 +12,26 @@ export const tabs = [
...
+    setSelectedTabId(tab.id);
+  };
+
   return (
     <div className="section">
       <h1 className="title">
-        Selected tab is Tab 1
+        Selected tab is
+        {' '}
+        {selectedTab.title}
```
dellachaise, line 27: Use [interpolation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals) here.

src/Tabs/Tabs.tsx
```diff
@@ -0,0 +1,52 @@
...
+    if (selectedTabId !== tab.id) {
+      onTabSelected(tab);
+    }
+  };
+
+  return (
+    <div data-cy="TabsComponent">
+      <div className="tabs is-boxed">
+        <ul>
+          {tabs.map((tab) => (
```
dellachaise, line 27: Use destructuring here.

dellachaise (changes requested):
Great job 👍 

Just add a few kittle changes according to comments 👇

src/Tabs/Tabs.tsx
ShmelyovOleg: I can't understand how to use destructuring here to then use tab in onclick

dellachaise: you could use this structure
```suggestion
      {tabs.map(tab => {
            const { id, title } = tab;
            return (
              your code here
            );
          })}
```

dellachaise (approved):
Well done 👍 

But still better use destructuring.

PR #1134, opened 2022-10-11
solution

MaxManis:
- [DEMO LINK](https://MaxManis.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,42 +1,31 @@
-import React from 'react';
+import React, { useState } from 'react';
+import { Tabs } from './components/Tabs';
+import { Tab } from './types/Tab';
+
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
 
-export const tabs = [
+export const tabs: Tab[] = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
   { id: 'tab-2', title: 'Tab 2', content: 'Some text 2' },
   { id: 'tab-3', title: 'Tab 3', content: 'Some text 3' },
 ];
 
 export const App: React.FC = () => {
+  const [selected, onTabSelected] = useState(tabs[0]);
+
   return (
     <div className="section">
       <h1 className="title">
-        Selected tab is Tab 1
+        {`Selected tab is ${selected.title}`}
       </h1>
 
-      <div data-cy="TabsComponent">
-        <div className="tabs is-boxed">
-          <ul>
-            <li className="is-active" data-cy="Tab">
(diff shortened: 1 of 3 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,60 @@
...
+            return (
+              <li
+                key={id}
+                className={classNames(
+                  selectedTabId === id
+                  || (
+                    !tabs.some((searchTab) => searchTab.id === selectedTabId)
+                    && index === 0
+                  )
+                    ? 'is-active' : '',
```
naziamoff, line 32: it is totally unclear from your code which style will be used here. Think of a way to make it readable
I would come up with creating separate variables for that

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,60 @@
...
+                </a>
+              </li>
+            );
+          })}
+        </ul>
+      </div>
+
+      <div className="block" data-cy="TabContent">
+        {
+          tabs.find((searchTab) => searchTab.id === selectedTabId)?.content
```
naziamoff, line 55: Better to wrap it into some relevant element

MaxManis: i changed `div.block` into article `p.block` ))

src/App.tsx
```diff
@@ -1,42 +1,31 @@
...
 
-export const tabs = [
+export const tabs: Tab[] = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
   { id: 'tab-2', title: 'Tab 2', content: 'Some text 2' },
   { id: 'tab-3', title: 'Tab 3', content: 'Some text 3' },
 ];
 
 export const App: React.FC = () => {
+  const [selected, onTabSelected] = useState(tabs[0]);
```
Anna-Tkachenko, line 16: what is selected? selected what? 🙂

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,58 @@
...
+      <div className="tabs is-boxed">
+        <ul>
+          {tabs.map((tab: Tab, index: number) => {
+            const { id, title } = tab;
+
+            return (
+              <li
+                key={id}
+                className={classNames({
+                  'is-active': selectedTabId === id || (!currentTab && !index),
```
Anna-Tkachenko, line 30: you repeated the check if current tab is selected 2 times
will be better to move this check to some variable before return statement in the map method, and use this variable where you need it

Anna-Tkachenko (changes requested):

Denchepornyuk (approved):

PR #1158, opened 2022-10-23
Implemented App and Tab components with tabs changing functionality

Dikey945:
[DEMO LINK](https://Dikey945.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
+import { Tabs } from './components/Tabs';
 
 export const tabs = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
@@ -10,33 +11,19 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [selectedTab, SetSelectedTab] = useState(tabs[0]);
+
   return (
     <div className="section">
       <h1 className="title">
-        Selected tab is Tab 1
+        {`Selected tab is ${tabs.find(tab => (
+          selectedTab.id === tab.id))?.title}`}
       </h1>
-
-      <div data-cy="TabsComponent">
-        <div className="tabs is-boxed">
-          <ul>
-            <li className="is-active" data-cy="Tab">
-              <a href="#tab-1" data-cy="TabLink">Tab 1</a>
-            </li>
-
-            <li data-cy="Tab">
-              <a href="#tab-2" data-cy="TabLink">Tab 2</a>
-            </li>
-
-            <li data-cy="Tab">
-              <a href="#tab-3" data-cy="TabLink">Tab 3</a>
-            </li>
-          </ul>
-        </div>
-
-        <div className="block" data-cy="TabContent">
-          Some text 1
-        </div>
-      </div>
+      <Tabs
+        tabs={tabs}
+        onTabSelected={SetSelectedTab}
+        selectedTabId={selectedTab.id}
+      />
     </div>
   );
 };
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -1,1 +1,61 @@
-export const Tabs = () => {};
+import React from 'react';
+import classNames from 'classnames';
+
+interface Tab {
+  id: string;
+  title: string;
+  content: string;
+}
+
+type Props = {
+  tabs: Tab[],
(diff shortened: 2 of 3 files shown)


src/App.tsx
```diff
@@ -10,33 +11,19 @@ export const tabs = [
...
 
 export const App: React.FC = () => {
+  const [selectedTab, SetSelectedTab] = useState(tabs[0]);
+
   return (
     <div className="section">
       <h1 className="title">
-        Selected tab is Tab 1
+        {`Selected tab is ${tabs.find(tab => (
+          selectedTab.id === tab.id))?.title}`}
```
ArturNagaycev, line 20: I think you can write like `{selectedTab.title}` without find method here

ArturNagaycev (approved):
GJ 👍

src/App.tsx
naziamoff: Yeah, looks like it is overcomplicated here

src/App.tsx
```diff
@@ -10,33 +11,19 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [selectedTab, SetSelectedTab] = useState(tabs[0]);
```
naziamoff, line 14: should the hook method be capitalized?

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,61 @@
...
+  selectedTabId: string;
+};
+
+export const Tabs: React.FC<Props> = ({
+  tabs,
+  onTabSelected,
+  selectedTabId,
+}) => {
+  const isSelectedTab = tabs.find((tab) => (
+    tab.id === selectedTabId)) || tabs[0];
```
naziamoff, line 22:
```suggestion
  const isSelectedTab = tabs
    .find((tab) => (tab.id === selectedTabId)) || tabs[0];
```

naziamoff: and should this actually be named `isSelectedTab`? Is it really boolean?

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,61 @@
...
+                  {title}
+                </a>
+              </li>
+            );
+          })}
+        </ul>
+      </div>
+
+      <article className="block" data-cy="TabContent">
+        {tabs.find(tab => (
```
naziamoff, line 56: DRY - Don't repeat yourself. You have already implemented this logic above

naziamoff (changes requested):

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,60 @@
...
+            return (
+              <li
+                className={classNames({ 'is-active': selectedTab })}
+                data-cy="Tab"
+                key={id}
+              >
+                <a
+                  href={`#${id}`}
+                  data-cy="TabLink"
+                  onClick={() => {
```
naziamoff, line 41: better to create a separate handler for that in order to improve readability

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,60 @@
...
+  const currentTab = tabs.find((tab) => (
+    tab.id === selectedTabId)) || tabs[0];
+
+  return (
+    <div data-cy="TabsComponent">
+      <div className="tabs is-boxed">
+        <ul>
+          {tabs.map((tab) => {
+            const { id, title } = tab;
+            const selectedTab = id === currentTab.id;
```
naziamoff, line 30: looks like a boolean variable to me, but is named just like I would name object 🤔
Imagine naming your dog Nazar. You will tell your friends about Nazar and they will think you're talking about a human, but it is actually a dog, so that name may easily confuse.

Naming matters!

naziamoff (changes requested):

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,61 @@
...
+      <div className="tabs is-boxed">
+        <ul>
+          {tabs.map((tab) => {
+            const { id, title } = tab;
+            const isSelectedTab = id === currentTab.id;
+            const handleClick = () => {
+              if (!isSelectedTab) {
+                onTabSelected(tab);
+              }
+            };
```
naziamoff, line 35: you should separate this logic

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,61 @@
...
+
+            return (
```
naziamoff, line 37: do you need second return here? Looks unclear to me

naziamoff (changes requested):

Anna-Tkachenko (approved):
⭐️

naziamoff (approved):

PR #1161, opened 2022-10-24
Solution

AMahalias:
[DEMO LINK](https://AMahalias.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,9 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
+import { Tabs } from './components/Tabs';
+import { Tab } from './types/Tab';
 
 export const tabs = [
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,7 +1,9 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
+import { Tabs } from './components/Tabs';
+import { Tab } from './types/Tab';
```
polosanya, line 6: Please add all files to your commit to fix this problem
image

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,50 @@
-export const Tabs = () => {};
+import classNames from 'classnames';
+import { Tab } from '../../types/Tab';
```
polosanya, line 2: Same problem here

polosanya (changes requested):
Good job! Please fix the problem with files
Probably that's the reason why you can't pass the tests

liliia-livandovskaya (approved):
Great job 🔥

PR #1162, opened 2022-10-24
react tabs was implemented

TheVovchik:
**[DEMO LINK](https://TheVovchik.github.io/react_tabs/)**

### DIFF
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -1,1 +1,48 @@
-export const Tabs = () => {};
+import classNames from 'classnames';
+import React from 'react';
+
+interface Props {
+  tabs: Tab[];
+  selectedTabId: string;
+  onTabSelected: (tab: Tab) => void;
+}
+
+export const Tabs: React.FC<Props> = ({
+  tabs,
+  selectedTabId,
+  onTabSelected,
+}) => {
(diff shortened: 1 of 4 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,48 @@
...
+}) => {
+  const selectedTab = tabs.find(tab => tab.id === selectedTabId) || tabs[0];
+
+  return (
+    <div data-cy="TabsComponent">
+      <div className="tabs is-boxed">
+        <ul>
+          {tabs.map(tab => (
+            <li
+              className={classNames({ 'is-active': tab.id === selectedTab.id })}
```
Anna-Borodaienko, line 23: You can add a variable for ```tab.id === selectedTab.id```

Anna-Borodaienko (approved):

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,52 @@
...
+                key={tab.id}
+              >
+                <a
+                  href={`#${tab.id}`}
+                  data-cy="TabLink"
+                  onClick={() => {
+                    return isSelected
+                      ? null
+                      : onTabSelected(tab);
+                  }}
```
naziamoff, line 37: we don't need to return null if the tab is not currently selected.
Try doing `isSelected && (...` - this should work. Or just usual `if`

naziamoff (approved):
🔥

PR #1165, opened 2022-10-24
react tabs was implemented

ArturNagaycev:
[DEMO LINK](https://ArturNagaycev.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,42 +1,32 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
 
-export const tabs = [
+import { Tabs } from './components/Tabs';
+import { Tab } from './react-app-env';
+
+export const tabs: Tab[] = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
   { id: 'tab-2', title: 'Tab 2', content: 'Some text 2' },
   { id: 'tab-3', title: 'Tab 3', content: 'Some text 3' },
 ];
 
 export const App: React.FC = () => {
+  const [selectedTab, setSelectedTab] = useState(tabs[0]);
+
   return (
     <div className="section">
       <h1 className="title">
-        Selected tab is Tab 1
+        {'Selected tab is '}
+        {selectedTab.title}
       </h1>
 
-      <div data-cy="TabsComponent">
-        <div className="tabs is-boxed">
-          <ul>
-            <li className="is-active" data-cy="Tab">
-              <a href="#tab-1" data-cy="TabLink">Tab 1</a>
-            </li>
-
-            <li data-cy="Tab">
-              <a href="#tab-2" data-cy="TabLink">Tab 2</a>
-            </li>
-
-            <li data-cy="Tab">
-              <a href="#tab-3" data-cy="TabLink">Tab 3</a>
-            </li>
-          </ul>
-        </div>
-
-        <div className="block" data-cy="TabContent">
-          Some text 1
-        </div>
-      </div>
+      <Tabs
+        tabs={tabs}
+        selectedTabId={selectedTab.id}
+        onTabSelected={setSelectedTab}
+      />
     </div>
   );
 };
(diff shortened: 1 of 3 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,52 @@
...
+  const rightActiveTab = tabs.some(tab => tab.id === selectedTabId)
+    ? selectedTabId
+    : tabs[0].id;
+
+  return (
+    <div data-cy="TabsComponent">
+      <div className="tabs is-boxed">
+        <ul>
+          {tabs.map(tab => (
+            <li
```
Dikey945, line 25: Better to destructure tab

Dikey945 (commented):
Good Job! 🔥

src/App.tsx
```diff
@@ -1,42 +1,32 @@
...
 
 export const App: React.FC = () => {
+  const [selectedTab, setSelectedTab] = useState(tabs[0]);
+
   return (
     <div className="section">
       <h1 className="title">
-        Selected tab is Tab 1
+        {'Selected tab is '}
+        {selectedTab.title}
```
naziamoff, line 22: do you think there's a way to combine them?

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,56 @@
...
+  selectedTabId: string,
+  onTabSelected: (tab: Tab) => void,
+};
+
+export const Tabs: React.FC<Props> = ({
+  tabs,
+  selectedTabId,
+  onTabSelected,
+}) => {
+  const rightActiveTab = tabs.some(tab => tab.id === selectedTabId)
```
naziamoff, line 16: not clear name of the variable, I can't get what is stored inside by its name. Any ideas to improve it?

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,56 @@
...
+                className={classNames({
+                  'is-active': rightActiveTab === id,
+                })}
+                data-cy="Tab"
+              >
+                <a
+                  href={`#${tab.id}`}
+                  data-cy="TabLink"
+                  onClick={() => {
+                    if (selectedTabId !== id) {
```
naziamoff, line 38: you should separate this logic to a handler

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,56 @@
...
+                  {title}
+                </a>
+              </li>
+            );
+          })}
+        </ul>
+      </div>
+
+      <div className="block" data-cy="TabContent">
+        {tabs.find(tab => selectedTabId === tab.id)?.content}
```
naziamoff, line 52: don't think that it is a good idea to look for a tab right where you render it.
You should create separate variable for that

naziamoff (changes requested):

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,61 @@
...
+  onTabSelected: (tab: Tab) => void,
+};
+type OnSelectTab = (id: string, tab: Tab) => void | undefined;
+
+export const Tabs: React.FC<Props> = ({
+  tabs,
+  selectedTabId,
+  onTabSelected,
+}) => {
+  const selectedTab = tabs.some(tab => tab.id === selectedTabId)
```
naziamoff, line 17: do you have to store exactly tabId here? I think just tab will do better

naziamoff: In that case you won't need selectedTabContent on line 26

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,61 @@
...
+
+  const selectedTabContent = tabs
+    .find(tab => selectedTabId === tab.id)?.content;
+
+  return (
+    <div data-cy="TabsComponent">
+      <div className="tabs is-boxed">
+        <ul>
+          {tabs.map(tab => {
+            const { id, title } = tab;
```
naziamoff, line 34: To me it seems that you don't really need destructuring here, those names are pretty short and clear. If you don't desctructure it  you'll have `tab.id`and `tab.title`, what doesn't look bad at all

naziamoff (approved):
GJ! But fixing comments is recommended

PR #1188, opened 2022-10-26
Implemented the `App` and `Tabs` components to render and switch betw…

AniaHna:
…een given `tabs`
[DEMO LINK](https://AniaHna.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
+import { Tabs } from './components/Tabs';
 
 export const tabs = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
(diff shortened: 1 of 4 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,51 @@
...
+              data-cy="Tab"
+              key={tab.id}
+            >
+              <a
+                href={`#${tab.id}`}
+                data-cy="TabLink"
+                onClick={() => {
+                  return tab.id !== selectedTab.id
+                    ? onTabSelected(tab)
+                    : null;
```
anton-iskryk, line 36: There's no need to use return here, add regular brackets instead curved ones. Also, you can use the operator `&&` to get rid of unnecessary null here.

anton-iskryk (commented):
Looks good to me. I wrote one comment.

anton-iskryk (approved):

src/App.tsx
```diff
@@ -10,33 +11,19 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [selectedTab, setSelectedTab] = useState(tabs[0]);
```
Denchepornyuk, line 14: add type for useState

src/components/Tabs/Tabs.tsx
Denchepornyuk: agree, you can get rid of return and null

Denchepornyuk (changes requested):

PR #1199, opened 2022-10-30
Develop

frontend112:
https://frontend112.github.io/react_tabs/

### DIFF
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -16,7 +16,7 @@ export const Tabs: FC<Props> = ({
   selectedTabId,
 }) => {
   const tabContent = tabs.find(tab => (
-    selectedTabId === tab.id
+    tab.id === selectedTabId
   ))?.content;
 
   const isIdCorrect:boolean = tabs.some((tab:Tab) => (
@@ -37,6 +37,7 @@ export const Tabs: FC<Props> = ({
                     {
                       'is-active':
                       selectedTabId === id
+                      // select first tab if selectedTabId is incorrect
                       || (!isIdCorrect && index === 0),
                     },
                   )}
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -8,14 +8,12 @@ type Props = {
   onTabSelected: (tabNr: Tab) => void,
   tabs: Tab[] | [],
   selectedTabId: string,
-  // selectedTab: Tab,
 };
 
 export const Tabs: FC<Props> = ({
   onTabSelected,
   tabs,
   selectedTabId,
-  // selectedTab,
 }) => {
   const tabContent = tabs.find(tab => (
     selectedTabId === tab.id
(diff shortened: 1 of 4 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,70 @@
...
+    tab.id === selectedTabId
+  ));
+
+  return (
+    <div className="TabsComponent" data-cy="tab-content">
+      <div className="tabs is-boxed">
+        <ul>
+          {
+            tabs.map((tab, index) => {
+              const { id, title } = tab;
```
jstmpelowycz, line 32:
```suggestion
            tabs.map(({ id, title }, index) => {
```

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,70 @@
...
+  tabs,
+  selectedTabId,
+}) => {
+  const tabContent = tabs.find(tab => (
+    tab.id === selectedTabId
+  ))?.content;
+
+  const isIdCorrect:boolean = tabs.some((tab:Tab) => (
+    tab.id === selectedTabId
+  ));
```
jstmpelowycz, line 24: no need to indicate type if it is too obvious (in this case – some returns boolean)
```suggestion
  const isIdCorrect = tabs.some((tab:Tab) => (
    tab.id === selectedTabId
  ));
```
also, if you indicate types, put a space in between
```
const someVariable: boolean = ...

// NOT the following way

const someVariable:boolean = ...
```

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,70 @@
-export const Tabs = () => {};
+import { FC } from 'react';
+
+import cn from 'classnames';
+
+import { Tab } from '../../types';
+
+type Props = {
+  onTabSelected: (tabNr: Tab) => void,
+  tabs: Tab[] | [],
```
jstmpelowycz, line 9: `tabs` will always be of type `Tab[]`
```suggestion
  tabs: Tab[],
```

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,70 @@
-export const Tabs = () => {};
+import { FC } from 'react';
+
+import cn from 'classnames';
+
+import { Tab } from '../../types';
+
+type Props = {
+  onTabSelected: (tabNr: Tab) => void,
```
jstmpelowycz, line 8: what does `tabNr` stand for? let's give our variables more meaningful names

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,70 @@
...
+              const { id, title } = tab;
+
+              return (
+                <li
+                  className={cn(
+                    {
+                      'is-active':
+                      selectedTabId === id
+                      // select first tab if selectedTabId is incorrect
+                      || (!isIdCorrect && index === 0),
```
jstmpelowycz, line 41: let's move this boolean to a separate local variable

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,70 @@
...
+                  key={id}
+                >
+                  <a
+                    href={`#${id}`}
+                    data-cy="TabLink"
+                    onClick={() => {
+                      if (id !== selectedTabId) {
+                        onTabSelected(tab);
+                      }
+                    }}
```
jstmpelowycz, line 54: do not write function right away inside props, create a separate handler instead

jstmpelowycz (changes requested):
awesome! let's apply a few fixes here

src/components/Tabs/Tabs.tsx
frontend112: @jstmpelowycz I don't think I can use this destructuring because I am passing tab object into onTabSelected below

jstmpelowycz (approved):
great job 🔥

PR #1202, opened 2022-10-31
solution

oleksandr-korobko:
[DEMO LINK](https://oleksandr-korobko.github.io/react_tabs/)

### DIFF
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -1,1 +1,49 @@
-export const Tabs = () => {};
+import React from 'react';
+import classNames from 'classnames';
+import { Tab } from '../../types/Tab';
+
+type Props = {
+  tabs: Tab[],
+  selectedTabId: string,
+  onTabSelected: (tab: Tab) => void,
+};
+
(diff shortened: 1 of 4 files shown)


avkovlit (approved):
LGFM!! Keep going!

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,49 @@
...
+  const selectedTab = tabs.find(({ id }) => id === selectedTabId) || tabs[0];
+
+  return (
+    <div data-cy="TabsComponent">
+      <div className="tabs is-boxed">
+        <ul>
+          {tabs.map(({ id, title, content }) => (
+            <li
+              key={id}
+              className={classNames({ 'is-active': id === selectedTab.id })}
```
Anna-Tkachenko, line 23: you check that current tab is selected several times in your code
let's create separate function for this check or just separate boolean variable inside the map callback and will use this variable where we need it

Anna-Tkachenko (changes requested):

Anna-Tkachenko (approved):
⭐️

PR #1219, opened 2022-11-17
Solution for tabs

sapnachoudhary06:
- [DEMO LINK](https://sapnachoudhary06.github.io/react_tabs/)

### DIFF
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -1,1 +1,57 @@
-export const Tabs = () => {};
+import classNames from 'classnames';
+import { Tab } from '../../types/Tab';
+
+type Props = {
+  tabs: Tab[];
+  selectedTabId: string;
+  onTabSelected: (tab: Tab) => void;
+};
+
+export const Tabs: React.FC<Props> = ({
+  tabs,
+  selectedTabId,
+  onTabSelected,
+}) => {
+  const clickHandler = (tab: Tab) => {
+    if (selectedTabId !== tab.id) {
+      onTabSelected(tab);
+    }
+  };
+
+  const selectedTab = tabs.find((tab) => (tab.id === selectedTabId)) || tabs[0];
+
+  return (
+    <div data-cy="TabsComponent">
+      <div className="tabs is-boxed">
+        <ul>
+          {tabs.map((tab) => (
+            <li
+              className={classNames(
+                {
+                  'is-active': tab.id === selectedTabId,
+                },
+              )}
+              data-cy="Tab"
+              key={tab.id}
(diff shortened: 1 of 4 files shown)


MeshackDev (approved):

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,57 @@
...
+  selectedTabId: string;
+  onTabSelected: (tab: Tab) => void;
+};
+
+export const Tabs: React.FC<Props> = ({
+  tabs,
+  selectedTabId,
+  onTabSelected,
+}) => {
+  const clickHandler = (tab: Tab) => {
```
ihor-jpeg, line 15: useCallback

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,57 @@
...
+    if (selectedTabId !== tab.id) {
+      onTabSelected(tab);
+    }
+  };
+
+  const selectedTab = tabs.find((tab) => (tab.id === selectedTabId)) || tabs[0];
```
ihor-jpeg, line 21: useMemo

ihor-jpeg (changes requested):

src/components/Tabs/Tabs.tsx
sapnachoudhary06: @ihor-jpeg, As I understand, `useCallback ` should be used for a method which is passed as a props to another component (so that the child component does not re-render each time a parent component is rendered and a new copy of method is created).

In the current scenario, the `clickHandler ` method is not being passed to any other component, so does it still add any value to use `useCallback`

src/components/Tabs/Tabs.tsx
sapnachoudhary06: @ihor-jpeg, As I understand, `useMemo` should be used for a variable which is passed as a props to another component (so that the child component does not re-render each time a parent component is rendered and a new copy of the variable is created).

In the current scenario, the `selectedTab` is not being passed to any other component, so does it still add any value to use `useMemo`

src/components/Tabs/Tabs.tsx
ihor-jpeg: yes, because the we re-render the component we don't want to create a new function in the memory every time

sapnachoudhary06: Done.

src/components/Tabs/Tabs.tsx
sapnachoudhary06: Done.

VinnieJ-2k20 (approved):

PR #1237, opened 2022-11-29
Render App with Tab selection

anastasiia-tilikina:
[DEMO LINK](https://anastasiia-tilikina.github.io/react_tabs/)

### DIFF
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -1,1 +1,58 @@
-export const Tabs = () => {};
+import React from 'react';
+import classNames from 'classnames';
+
+export type Tab = {
+  id: string;
+  title: string;
+  content: string;
+};
+
+type Props = {
+  tabs: Tab[]
+  selectedTabID: string
+  onTabSelected: (tab: Tab) => void
+};
+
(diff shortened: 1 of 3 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,58 @@
-export const Tabs = () => {};
+import React from 'react';
+import classNames from 'classnames';
+
+export type Tab = {
```
vlad-demchuk, line 4: you can avoid exporting if you use a type in the same file

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,58 @@
...
+            >
+              <a
+                data-cy="TabLink"
+                href={`#${tab.id}`}
+                onClick={() => {
+                  if (selectedTab.id === tab.id) {
+                    return;
+                  }
+
+                  onTabSelected(tab);
```
tetiana-ivanchenko, line 43: move it in the separate method

tetiana-ivanchenko (changes requested):
GJ, just a small improvement

nazarmatsevych (approved):

PR #1242, opened 2022-11-30
Solution

bokorzita:
https://bokorzita.github.io/react_tabs/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import React, { useState } from 'react';
+import { Tabs } from './components/Tabs';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -10,33 +11,27 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [selectedTab, setSelectedTab] = useState(tabs[0]);
+
+  const onTabSelected = (tabId: string) => {
+    const selected = tabs.find((tab) => tab.id === tabId);
```
polosanya, line 17:
```suggestion
    const selected = tabs.find((tab) => tab.id === tabId) || tabs[0];
```
image

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,36 @@
...
+          <li
+            key={tab.id}
+            data-cy="Tab"
+            className={tab === selectedTab ? 'is-active' : ''}
+          >
+            <a
+              href={`#${tab.id}`}
+              data-cy="TabLink"
+              onClick={() => onTabSelected(tab.id)}
+              onKeyDown={() => onTabSelected(tab.id)}
```
polosanya, line 23: Do you really need it? It works with Enter by default, no need to add all the keys

polosanya (changes requested):
Good job! Please check the comments and let's try to pass the tests)
Feel free to ask for help in the chat if you have any difficulties

PR #1279, opened 2022-12-09
React tabs task solution

juliiaap:
- [DEMO LINK](https://juliiaap.github.io/react_tabs/)

### DIFF
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -1,1 +1,58 @@
-export const Tabs = () => {};
+import classNames from 'classnames';
+
+export interface Tab {
+  id: string,
+  title: string,
+  content: string,
+}
+
+type Props = {
+  tabs: Tab[];
+  selectedTabId: string;
+  onTabSelected: CallableFunction;
+};
+
+export const Tabs: React.FC<Props> = ({
+  tabs,
+  selectedTabId,
+  onTabSelected,
+}) => {
+  return (
+    <div data-cy="TabsComponent">
+      <div className="tabs is-boxed">
+        <ul>
+          {tabs.map(tab => (
+            <li
+              className={classNames({ 'is-active': tab.id === selectedTabId })}
+              data-cy="Tab"
+            >
+              <a
+                href={`#${tab.id}`}
+                data-cy="TabLink"
+                onClick={(event) => {
+                  if (event.currentTarget.hash.slice(1) === selectedTabId) {
+                    return;
+                  }
+
+                  const currentTab = {
+                    id: event.currentTarget.hash.slice(1),
+                    title: event.currentTarget.textContent,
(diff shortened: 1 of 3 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,58 @@
...
+  onTabSelected,
+}) => {
+  return (
+    <div data-cy="TabsComponent">
+      <div className="tabs is-boxed">
+        <ul>
+          {tabs.map(tab => (
+            <li
+              className={classNames({ 'is-active': tab.id === selectedTabId })}
+              data-cy="Tab"
```
polosanya, line 27: Add `key` attribute
https://www.developerway.com/posts/react-key-attribute

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,58 @@
...
+            >
+              <a
+                href={`#${tab.id}`}
+                data-cy="TabLink"
+                onClick={(event) => {
+                  if (event.currentTarget.hash.slice(1) === selectedTabId) {
```
polosanya, line 33: Move it to the separate function

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,58 @@
...
+              >
+                {tab.title}
+              </a>
+            </li>
+          ))}
+        </ul>
+      </div>
+
+      <div className="block" data-cy="TabContent">
+        {tabs.find(tab => tab.id === selectedTabId)?.content}
```
polosanya, line 54:
```suggestion
        {tabs.find(tab => tab.id === selectedTabId)?.content || tabs[0].content}
```
Let's try this. Maybe it will help to pass the test

polosanya (changes requested):
Almost done! Let's provide some fixes

src/components/Tabs/Tabs.tsx
juliiaap: fixed

src/components/Tabs/Tabs.tsx
juliiaap: done

src/components/Tabs/Tabs.tsx
juliiaap: I changed it, but unfortunately, it didn't help to pass the test.

loralevitska: Hi! Try to put it in a variable and use hooks to update the selected tab in the state, and use only the variable here

loralevitska (approved):
good job!

src/components/Tabs/Tabs.tsx
juliiaap: Hi! It works with hooks. Thanks. But I didn't use them in the previous solution because there is a condition in the task

> The `Tabs` component should be stateless (don't have internal state, only props).

PR #1280, opened 2022-12-13
solution

MelnychukYana:
https://MelnychukYana.github.io/react_tabs/


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,55 @@
...
+  onTabSelected: (tab: TabType) => void,
+  selectedTabId: string,
+}
+
+export const Tabs: React.FC<Props> = ({
+  tabs,
+  onTabSelected,
+  selectedTabId,
+}) => {
+  const selectedTab = tabs.find(tab => tab.id === selectedTabId) || tabs[0];
```
Bohdan-Maksymiuk, line 15: you found it twice
so in this case better to store whole tab in state

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,55 @@
...
+      <div className="tabs is-boxed">
+        <ul>
+          {
+            tabs.map(tab => {
+              return (
+                <li
+                  className={classNames(
+                    { 'is-active': selectedTab.id === tab.id },
+                  )}
+                  data-cy="Tab"
```
Bohdan-Maksymiuk, line 34: where is [key](https://reactjs.org/docs/lists-and-keys.html)

Bohdan-Maksymiuk (changes requested):

PR #1284, opened 2022-12-17
Tabs

s3rhiienko:
[DEMO LINK](https://s3rhiienko.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
+import { Tab, Tabs } from './components/Tabs';
 
 export const tabs = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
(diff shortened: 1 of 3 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,79 @@
...
+
+  useEffect(() => {
+    const isActiveTab = tabs.find(tab => tab.id === selectedTabId)?.content
+      || tabs[0].content;
+    const isActiveTabId = tabs.find(tab => tab.id === selectedTabId)?.id
+      || tabs[0].id;
+
+    setActiveTab(isActiveTab);
+    setActiveTabId(isActiveTabId);
+  }, [selectedTabId]);
```
SanyaBratashchuk, line 48: simpify it. Pass tabs array as a prop, and on click chage selected tab than when you map you array if current id is an id of selectedTab show its content

src/App.tsx
```diff
@@ -10,33 +11,25 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [selectedTab, setSelectedTab] = useState(tabs[0].id);
```
SanyaBratashchuk, line 14:
```suggestion
  const [selectedTabId, setSelectedTabId] = useState(tabs[0].id);
```

SanyaBratashchuk (changes requested):

PR #1288, opened 2022-12-23
done

Nazar-Brunarskyi:
[DEMO LINK](https://Nazar-Brunarskyi.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,10 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
+import { Tabs } from './components/Tabs';
+import { Tab } from './types/Tab';
+import { getInfoFromTab } from './functions/getInfoFromTab';
 
 export const tabs = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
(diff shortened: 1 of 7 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,58 @@
...
+    : tabs[0].id;
+
+  return (
+    <div data-cy="TabsComponent">
+      <div className="tabs is-boxed">
+        <ul>
+          {
+            tabs.map(tab => (
+              <li
+                className={classNames({ 'is-active': isActive === tab.id })}
```
romanblyashuk94, line 30: when variable name starts from "is" - it means that this variable contains boolean value
so you can name it "activeTabId"

but i can suggest better way

```suggestion
	const selectedTab = tabs.find(({ id }) => id === selectedTabId) || tabs[0];
	
  return (
    <div data-cy="TabsComponent">
      <div className="tabs is-boxed">
        <ul>
          {
            tabs.map(tab => (
              <li
                className={classNames({ 'is-active': selectedTab.id === tab.id })}
```

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,58 @@
...
+                key={tab.id}
+              >
+                <a
+                  href={`#${tab.id}`}
+                  data-cy="TabLink"
+                  onClick={() => {
+                    if (selectedTabId !== tab.id) {
+                      onTabSelected(tab);
+                    }
+                  }}
```
romanblyashuk94, line 41: when callback just call other function - you can leave it in murkup
but when it contain some logic, it will be better to move it from markup to some variable
```suggestion
const handleTabClick = (id: number) => {
	if (selectedTabId !== tab.id) {
		onTabSelected(tab);
	}
}
	...	
  onClick={() => handleTabClick(tab.id)}
```

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,58 @@
...
+                </a>
+              </li>
+            ))
+          }
+        </ul>
+      </div>
+
+      <div className="block" data-cy="TabContent">
+        {
+          getInfoFromTab(selectedTabId, 'content', tabs) || tabs[0].content
```
romanblyashuk94, line 53: if you'll use my wroted higher suggest , you can leave here just:
```suggestion
          {selectedTab.content}
```

romanblyashuk94 (changes requested):

romanblyashuk94 (approved):
Good job 👍

src/App.tsx
```diff
@@ -10,33 +13,25 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [selectedTabId, setSelectedTabId] = useState('tab-1');
```
valerii-derkach, line 16:
```suggestion
  const [selectedTabId, setSelectedTabId] = useState('tab-1');
```

maybe better use tabs[0].id

valerii-derkach (approved):

PR #1303, opened 2023-01-04
solution

lefteris-L:
[DEMO LINK](https://lefteris-l.github.io/react_tabs/)

### DIFF
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -5,52 +5,58 @@ import { Tab } from '../../types';
 type Props = {
   tabs: Tab[]
   selectedTabId: string
-  content: string
   onTabSelected: (tab: Tab) => void
 };
 
 export const Tabs: React.FC<Props> = ({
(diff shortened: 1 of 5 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,62 @@
...
+  const { content } = selectedTab;
+
+  return (
+    <div data-cy="TabsComponent">
+      <div className="tabs is-boxed">
+        <ul>
+          {tabs.map(tab => {
+            const { id, title } = tab;
+
+            return (
```
meforoma, line 29: could be a bit shorter
```suggestion
          {tabs.map(({ id, title }) => (
```

meforoma (approved):
✅

src/components/Tabs/Tabs.tsx
lefteris-L: That's what I did at first, but I needed `tab` to use in `onClick`

PR #1304, opened 2023-01-05
solution

serhii-yunnikov:
https://serhii-yunnikov.github.io/react_tabs/

### DIFF
--- a/src/types/tab.ts
+++ b/src/types/tab.ts
@@ -1,0 +1,5 @@
+export interface Tab {
+  id: string,
+  title: string,
+  content: string,
+}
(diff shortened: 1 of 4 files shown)


src/types/tab.ts
```diff
@@ -0,0 +1,5 @@
+export interface Tab {
+  id: string,
+  title: string,
+  content: string,
+}
```
marina-tilniak, line 5: tip: you can use **react-app-env.d.ts** file for types and in this case you don't need export/import

marina-tilniak (approved):

PR #1315, opened 2023-01-08
Develop

IoghiStudio:
[DEMO LINK](https://IoghiStudio.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,42 +1,34 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
+import { Tab } from './types/tab';
+import { Tabs } from './components/Tabs/Tabs';
 
-export const tabs = [
+export const tabs: Tab[] = [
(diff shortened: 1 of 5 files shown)


src/App.tsx
```diff
@@ -1,42 +1,34 @@
...
 
-export const tabs = [
+export const tabs: Tab[] = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
   { id: 'tab-2', title: 'Tab 2', content: 'Some text 2' },
   { id: 'tab-3', title: 'Tab 3', content: 'Some text 3' },
 ];
 
 export const App: React.FC = () => {
+  const [selectedTab, setTab] = useState(tabs[0]);
```
nurdanokcu, line 15:
```suggestion
  const [selectedTab, setSelectedTab] = useState(tabs[0]);
```
It would make more sense to name it this way. It is a small project there is only one hook but it would matter for bigger projects.

nurdanokcu (approved):
Approving it with a naming suggestion.

src/App.tsx
naziamoff: I agree with @nurdanokcu on that one

naziamoff (approved):

PR #1317, opened 2023-01-09
Develop

nurdanokcu:
[DEMO LINK](https://nurdanokcu.github.io/react_tabs/)

### DIFF
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -1,1 +1,58 @@
-export const Tabs = () => {};
+import cn from 'classnames';
+import { Tab } from '../../types/Tab';
+
+type Props = {
+  tabs: Tab[];
+  selectedTabId: string;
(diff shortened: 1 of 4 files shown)


UmizDemud (approved):
I'm looking for an obvious mistake, unaware that it is your code 😄 

👍 Looks good to me

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,58 @@
...
+                    if (id !== selectedTabId) {
+                      onTabSelected(tab);
+                    }
+                  }}
+                >
+                  {title}
+                </a>
+              </li>
+            );
+          })}
```
meforoma, line 46: i think this would deserve it's own `.tsx` component file
It's simply looks kind of bulky right inside the return statement
Although it's not that big so we can live with that 😉

meforoma (approved):
traditional 👍🏼

PR #1321, opened 2023-01-11
Solution

Anastasia4sik:
[DEMO LINK](https://Anastasia4sik.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,9 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
+import { Tabs } from './components/Tabs';
+import { Tab } from './types/tab';
 
 export const tabs = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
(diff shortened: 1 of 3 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,57 @@
...
+    onTabSelected,
+  } = props;
+
+  const handleSelectedTab = (tab: Tab) => {
+    if (tab.id === selectedTabId) {
+      return;
+    }
+
+    onTabSelected(tab);
+  };
```
kreisberg, line 24: `
 if (tab.id !== selectedTabId) {
       selectedOnTab(tab);
     }
`

here we don't need to use two conditions

kreisberg (approved):
Find only one mistakes with conditions. BTW everything seems great! 🌹

ivan-kadykalo (changes requested):
Solution looks good, but demo link doesn't work
image

src/App.tsx
```diff
@@ -10,33 +12,23 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [selectedTab, hasSelectedTab] = useState(tabs[0]);
```
valerii-derkach, line 15: naming! [name, setName]

valerii-derkach (approved):

PR #1351, opened 2023-01-12
add interface Tab and markup for App component

kreisberg:


src/App.tsx
```diff
@@ -1,42 +1,77 @@
...
+import { Tab } from './components/Tabs/types/Tab';
 
-export const tabs = [
-  { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
-  { id: 'tab-2', title: 'Tab 2', content: 'Some text 2' },
-  { id: 'tab-3', title: 'Tab 3', content: 'Some text 3' },
+export const tabs: Tab[] = [
+  {
+    id:
+    'tab-1',
```
anastasiia-bilska, line 12: This needs to be in one line

src/App.tsx
```diff
@@ -1,42 +1,77 @@
...
 
-        <div className="block" data-cy="TabContent">
-          Some text 1
-        </div>
-      </div>
+      <img
+        src=""
+        alt=""
+        className={`cat${selectedTab.cat}`}
+      />
```
anastasiia-bilska, line 74: I think it is better to use div for background image. I am not sure, but is seems that your images are broken, because your tag  doesn't have src. + If you use , you need to specify alt for better SEO.

anastasiia-bilska: image

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,55 @@
...
+  selectedTabId: string,
+  selectedOnTab: (tab: Tab) => void,
+};
+
+export const Tabs: React.FC<Props> = ({
+  tabs,
+  selectedTabId = tabs[0],
+  selectedOnTab,
+}) => {
+  const selectedTab = tabs.find(({ id }) => id === selectedTabId);
```
anastasiia-bilska, line 15: find can return undefind. Please add to `selected tab` other option for this case

anastasiia-bilska (changes requested):
Good, check my coments, please!

src/components/Tabs/Tabs.tsx
kreisberg: When I tried to add it, the picture breaks. Perhaps the reason is the small weight of the images 🤔

src/App.tsx
kreisberg: When I tried to add it, the picture breaks. Perhaps the reason is the small weight of the images 🤔

src/App.tsx
kreisberg: Fix it!

src/components/Tabs/Tabs.tsx
kreisberg: In progress, looking for another way 🤔

src/App.tsx
```diff
@@ -1,42 +1,76 @@
...
+  },
 ];
 
 export const App: React.FC = () => {
-  return (
-    <div className="section">
-      <h1 className="title">
-        Selected tab is Tab 1
-      </h1>
+  const [darkMode, setDarkMode] = useState(false);
```
valerii-derkach, line 31:
```suggestion
  const [isDarkMode, setDarkMode] = useState(false);
```

src/components/Tabs/Tabs.tsx
valerii-derkach, line 15:
```suggestion
  const selectedTab = tabs.find(({ id }) => id === selectedTabId) || tabs[0];
```

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,55 @@
...
+              >
+                {tab.title}
+              </a>
+            </li>
+          ))}
+        </ul>
+      </div>
+
+      <div className="block" data-cy="TabContent">
+        {selectedTab?.content}
```
valerii-derkach, line 50:
```suggestion
        {selectedTab.content}
```

src/App.tsx
```diff
@@ -1,42 +1,76 @@
...
 
-        <div className="block" data-cy="TabContent">
-          Some text 1
-        </div>
-      </div>
+      <img
+        src=""
+        alt=""
+        className={`cat${selectedTab.cat}`}
+      />
```
valerii-derkach, line 73: try to add your pictures to `src` instead of `bg-images` and fill `alt` attribute

valerii-derkach (changes requested):
Good job!

small fixes

src/App.tsx
kreisberg: Use div and background-image and it's working ✔️

src/components/Tabs/Tabs.tsx
kreisberg: I tried to delete it, but compiler gives an error.  Can't deploy without '?'..

src/components/Tabs/Tabs.tsx
kreisberg: I added first page (tabs[0]) as default value in properties ✔️

So if nothing was selected, we always see the first tab.

src/App.tsx
```diff
@@ -1,42 +1,81 @@
...
+    id: 'tab-1',
+    title: 'Tab 1',
+    content: 'First week with React. BTW the second too.',
+    cat: 1,
+  },
+  {
+    id: 'tab-2',
+    title: 'Tab 2',
+    content: 'Ihor said he was upset. OMG whats wrong with us, dudes.',
+    cat: 2,
```
ivan-kadykalo, line 20: Add image links or names here, and set them like a props

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,55 @@
...
+                  handleOnClick(tab);
+                }}
+              >
+                {tab.title}
+              </a>
+            </li>
+          ))}
+        </ul>
+      </div>
+
```
ivan-kadykalo, line 48: Use images with tags

src/App.scss
```diff
@@ -1,3 +1,139 @@
...
+
+.cat1 {
+  background-image: url("cat1.jpg");
+  background-repeat: no-repeat;
+
+  position: fixed;
+  top: 246px;
+  left: 45px;
+  width: 100%;
+  height: 100%;
```
ivan-kadykalo, line 105: Each style should be in the corresponding file next to its component.

src/App.scss
```diff
@@ -1,3 +1,139 @@
...
+
+.cat2 {
+  background-image: url("cat2.jpg");
+  background-repeat: no-repeat;
+
+  position: fixed;
+  top: 246px;
+  left: 45px;
+  width: 100%;
+  height: 100%;
```
ivan-kadykalo, line 116: Duplicated styles

src/App.scss
```diff
@@ -1,3 +1,120 @@
...
+
+.cat {
+  background-repeat: no-repeat;
+  position: fixed;
+  top: 246px;
+  left: 45px;
+  width: 100%;
+  height: 100%;
+
+  &1 {
```
ivan-kadykalo, line 105:
```suggestion
  &--1 {
```

ivan-kadykalo (approved):
Good job 🔥

PR #1365, opened 2023-01-20
Solution

Grayfox147:
[DEMO LINK](https://Grayfox147.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
+import { Tabs } from './components/Tabs';
 
 export const tabs = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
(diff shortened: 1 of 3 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,58 @@
...
+                })}
+                data-cy="Tab"
+              >
+                <a
+                  href={`#${tab.id}`}
+                  data-cy="TabLink"
+                  onClick={() => {
+                    if (tab.id !== isSelectedTab.id) {
+                      onTabSelected(tab.id);
+                    }
```
CharAnagn, line 43: Here is the problem with the test. It requires you to pass the tab as an argument.

CharAnagn (changes requested):
After you make some adjustments it should work. If you need a hint, comment here but i think you got it 💯

src/components/Tabs/Tabs.tsx
Grayfox147: I am not sure, so you are saying that instead of passing tab.id I should pass the whole object tab, somehow? is this correct?

CharAnagn: Yup you found it 👍

CharAnagn (approved):
GG 🔥

src/App.tsx
```diff
@@ -10,32 +11,27 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [selectedTabId, setSelectedTabId] = useState('tab-1');
+  const isSelectedTab = tabs.find((tab) => selectedTabId === tab.id)
+    || tabs[0];
```
naziamoff, line 16: why not just store `selectedTab` here, you're passing whole object to the handler anyway. This way you won't need to do `tabs.find` in both components

naziamoff (approved):
gj, just one suggestion

PR #1366, opened 2023-01-20
Solution

CharAnagn:
[DEMO LINK](https://charanagn.github.io/react_tabs/)

If you name the selectedTabId prop that you pass on Tabs other than that name, 3 tests are failing.  It took me a some time to figure out what's wrong 🥲

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,42 +1,37 @@
-import React from 'react';
+/* eslint-disable no-console */
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
+import { Tab } from './components/types/types';
+import { Tabs } from './components/Tabs';
 
-export const tabs = [
+export const tabs: Tab[] = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,42 +1,37 @@
...
 
-export const tabs = [
+export const tabs: Tab[] = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
   { id: 'tab-2', title: 'Tab 2', content: 'Some text 2' },
   { id: 'tab-3', title: 'Tab 3', content: 'Some text 3' },
 ];
 
 export const App: React.FC = () => {
+  const [selectedTabId, setSelectedTab] = useState(tabs[0].id);
```
naziamoff, line 16: you can store the whole tab here, this way you won't need to repeat the logic of `find` in both components

src/App.tsx
```diff
@@ -1,42 +1,37 @@
...
+
+  const tabSelectHandler = (tab: Tab) => {
```
naziamoff, line 18:
```suggestion
  const selectTab = (tab: Tab) => {
```
or 
```suggestion
  const handleTabSelect = (tab: Tab) => {
```

naziamoff (approved):
🔥

PR #1369, opened 2023-01-22
solution

mmahfuzi:
- [DEMO LINK](https://mmahfuzi.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
+import { Tabs } from './components/Tabs/Tabs';
 
 export const tabs = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
@@ -10,33 +11,24 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [selectedTabId, setSelectedTabId] = useState('tab-1');
+  const [selectedTabTitle, setSelectedTabTitle] = useState('Tab 1');
+  const [selectedTabContent, setSelectedTabContent] = useState('Some text 1');
+
+  const selectedTab = (id: string, title: string, content: string) => {
+    setSelectedTabId(id);
+    setSelectedTabTitle(title);
+    setSelectedTabContent(content);
+  };
+
   return (
-    <div className="section">
-      <h1 className="title">
-        Selected tab is Tab 1
-      </h1>
+    <Tabs
+      tabs={tabs}
+      selectedTabTitle={selectedTabTitle}
+      selectedTabId={selectedTabId}
+      selectedTabContent={selectedTabContent}
+      onSelectedTab={selectedTab}
 
-      <div data-cy="TabsComponent">
-        <div className="tabs is-boxed">
-          <ul>
-            <li className="is-active" data-cy="Tab">
-              <a href="#tab-1" data-cy="TabLink">Tab 1</a>
-            </li>
-
-            <li data-cy="Tab">
-              <a href="#tab-2" data-cy="TabLink">Tab 2</a>
-            </li>
-
-            <li data-cy="Tab">
-              <a href="#tab-3" data-cy="TabLink">Tab 3</a>
-            </li>
-          </ul>
-        </div>
-
-        <div className="block" data-cy="TabContent">
-          Some text 1
-        </div>
-      </div>
-    </div>
+    />
   );
 };
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -1,1 +1,59 @@
-export const Tabs = () => {};
+import React from 'react';
+import cn from 'classnames';
+
+interface Tab {
+  id: string,
+  title: string,
+  content: string,
+}
+
+type Props = {
+  tabs: Tab[];
+  selectedTabTitle: string;
+  selectedTabId: string;
+  selectedTabContent: string,
+  onSelectedTab: (id: string, title: string, content: string) => void,
+
+};
+
+export const Tabs:React.FC<Props> = ({
+  tabs,
+  selectedTabTitle,
+  selectedTabId,
+  selectedTabContent,
+  onSelectedTab,
+
+}) => (
+  <div className="section">
+    <h1 className="title">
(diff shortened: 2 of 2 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,59 @@
...
+  onSelectedTab,
+
+}) => (
+  <div className="section">
+    <h1 className="title">
+      {`Selected tab is ${selectedTabTitle}`}
+    </h1>
+
+    <div data-cy="TabsComponent">
+
```
CharAnagn, line 33: ( I will comment in-order with the steps that will help you also pass the tests)

Keep the div with class section in the App component and move only the block with the data-cy="TabsComponent".

src/App.tsx
```diff
@@ -10,33 +11,24 @@ export const tabs = [
...
-
-      <div data-cy="TabsComponent">
-        <div className="tabs is-boxed">
-          <ul>
-            <li className="is-active" data-cy="Tab">
-              <a href="#tab-1" data-cy="TabLink">Tab 1</a>
-            </li>
+  const [selectedTabId, setSelectedTabId] = useState('tab-1');
+  const [selectedTabTitle, setSelectedTabTitle] = useState('Tab 1');
+  const [selectedTabContent, setSelectedTabContent] = useState('Some text 1');
```
CharAnagn, line 16: Keep only one state, the selectedTabId and use the find method to save the selectedTab based on the id

src/App.tsx
```diff
@@ -10,33 +11,24 @@ export const tabs = [
...
 
-            <li data-cy="Tab">
-              <a href="#tab-2" data-cy="TabLink">Tab 2</a>
-            </li>
+  const selectedTab = (id: string, title: string, content: string) => {
+    setSelectedTabId(id);
+    setSelectedTabTitle(title);
+    setSelectedTabContent(content);
+  };
```
CharAnagn, line 22: 👍 But now you will accept a tab from your child component, so you can use your setSelectedTabId accordingly. (It is also required for one of the tests to get a tab from the child instead of tab.id for example)

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,59 @@
...
+          {tabs.map(tab => (
+            <li
+              className={cn({ 'is-active': selectedTabId === tab.id })}
+              data-cy="Tab"
+            >
+              <a
+                href={`#${tab.id}`}
+                data-cy="TabLink"
+                onClick={() => onSelectedTab(tab.id, tab.title, tab.content)}
+              >
```
CharAnagn, line 45: Here is for my previous comment, that you need to pass the tab as argument.

CharAnagn (changes requested):
Overall you got the idea you just need to pass the tests ( took my a lot of time especially with one 🥲). 

There are still some changes you need to make to pass some others that i didn't comment, I don't want to spoil all the fun, but once you pass all the tests for the Page it will be clear what you need to do for the Tab component. If you need help with any tests comment here the test. Have fun 🙂

src/App.tsx
mmahfuzi: I implemented using selectedTab as a state. To avoid calling find in the runtime

CharAnagn: True I agree, I missunderstood the instructions, I thought they wanted us to use only one state.

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,50 @@
...
+export const Tabs:React.FC<Props> = ({
+  tabs,
+  selectedTabId,
+  onSelectTab,
+}) => {
+  let selectedTabIndex = tabs.findIndex(tab => tab.id === selectedTabId);
+
+  selectedTabIndex = selectedTabIndex !== -1
+    ? selectedTabIndex
+    : 0;
```
meforoma, line 20: @mmahfuzi you can skip the index step - just apply the `.find` method
and then if it's not found, apply `tabs[0]`
```suggestion
  let selectedTab = tabs.find(tab => tab.id === selectedTabId)
    || tabs[0];
```

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,50 @@
...
+                {tab.title}
+              </a>
+            </li>
+          ))}
+
+        </ul>
+      </div>
+
+      <div className="block" data-cy="TabContent">
+        {tabs[selectedTabIndex].content}
```
meforoma, line 46:
```suggestion
        {selectedTab.content}
```

meforoma (changes requested):
great job @mmahfuzi 👏🏼 
- i've left one comment to clear
- also it seems relatively easy to pass all the tests
  - refer to the test file here: https://github.com/mmahfuzi/react_tabs/blob/develop/src/components/Tabs/Tabs.spec.jsx

- and make sure to cover the failing tests
  - i think 3 of them will be covered by renaming the method to `onTabSelected`
failing tests

mmahfuzi:
> great job @mmahfuzi 👏🏼
> 
> * i've left one comment to clear
> * also it seems relatively easy to pass all the tests
>   
>   * refer to the test file here: https://github.com/mmahfuzi/react_tabs/blob/develop/src/components/Tabs/Tabs.spec.jsx
> * and make sure to cover the failing tests
>   
>   * i think 3 of them will be covered by renaming the method to `onTabSelected`
> 
> failing tests

Hello @meforoma thank you :)

I applied the changes you requested but the npm test doesn't work!

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,47 @@
-export const Tabs = () => {};
+import React from 'react';
+import cn from 'classnames';
+import { Tab } from '../../types/Tab';
+
+type Props = {
+  tabs: Tab[];
+  selectedTabId: string;
+  onSelectTab: (tab: Tab) => void;
```
meforoma, line 8:
```suggestion
  onTabSelected: (tab: Tab) => void;
```

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,47 @@
...
+};
+
+export const Tabs:React.FC<Props> = ({
+  tabs,
+  selectedTabId,
+  onSelectTab: onTabSelected,
```
meforoma, line 14: onTabSelected

meforoma, line 14:
```suggestion
  onTabSelected,
```

meforoma (approved):
this will do ✅ 
well done 👍🏼

PR #1371, opened 2023-01-22
Develop

kenvastolord:
[DEMO LINK](https://kenvastolord.github.io/react_tabs/)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,52 @@
...
+            className={classNames({ 'is-active': tab.id === selectedTabId })}
+            data-cy="Tab"
+            key={tab.id}
+          >
+            <a
+              href={`#${tab.id}`}
+              data-cy="TabLink"
+              onClick={() => {
+                if (tab.id !== selectedTabId) {
+                  onTabSelected(tab.id);
```
CharAnagn, line 35: Here the tests require to pass the tab as an argument and not the tab.id

src/App.tsx
```diff
@@ -10,33 +11,21 @@ export const tabs = [
...
 export const App: React.FC = () => {
+  const [selectedTabId, setSelectedTabId] = useState('tab-1');
+
   return (
     <div className="section">
       <h1 className="title">
-        Selected tab is Tab 1
+        {tabs.map((tab) => (
+          tab.id === selectedTabId && (`Selected tab is ${tab.title}`)
+        ))}
```
CharAnagn, line 21: I would suggest to create a variable and use the find method to find the tab that you are looking for based on the id.

CharAnagn (changes requested):
For the second test that you are failing, if you look at the tests it passes a selectedTabId to your Tabs component that doesn't match with any tab. So you need to find a way that you'll search for the tab that has the selectedTabId and if you don't find one, instead of returning undefined you should select the first tab.

src/App.tsx
```diff
@@ -1,42 +1,31 @@
...
+export const tabs: Tab[] = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
   { id: 'tab-2', title: 'Tab 2', content: 'Some text 2' },
   { id: 'tab-3', title: 'Tab 3', content: 'Some text 3' },
 ];
 
 export const App: React.FC = () => {
+  const [selectedTabId, setSelectedTabId] = useState('tab-1');
+  const getTabById = tabs.find(tab => tab.id === selectedTabId);
+  const handleSelectedTab = (tab: Tab) => setSelectedTabId(tab.id);
```
naziamoff, line 16: you are passing the tab to this component anyway. So there's no need to get the tab by id, you can just store the whole tab to your state

naziamoff (approved):
well done. Just one suggestion

PR #1372, opened 2023-01-23
Solution

Rijo-J:
- [DEMO LINK](https://rijo-j.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,31 +1,36 @@
-import React, { useState } from 'react';
+/* eslint-disable no-console */
+import { FC, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
 import { Tabs } from './components/Tabs';
 import { Tab } from './types/Tab';
 
-export const tabs = [
+export const tabs: Tab[] = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
   { id: 'tab-2', title: 'Tab 2', content: 'Some text 2' },
   { id: 'tab-3', title: 'Tab 3', content: 'Some text 3' },
 ];
 
-export const App: React.FC = () => {
-  const [selectedTab, setSelectedTab] = useState(tabs[0]);
+export const App: FC = () => {
+  const [selectedTabId, setSelectedTab] = useState(tabs[0].id);
 
-  const handleTabClick = (tab: Tab) => setSelectedTab(tab);
+  const tabSelectHandler = (tab: Tab) => {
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,42 +1,37 @@
...
   { id: 'tab-3', title: 'Tab 3', content: 'Some text 3' },
 ];
 
-export const App: React.FC = () => {
+export const App: FC = () => {
+  const [selectedTabId, setSelectedTab] = useState(tabs[0].id);
+
+  const tabSelectHandler = (tab: Tab) => {
+    setSelectedTab(tab.id);
+  };
```
kenvastolord, line 20: Great work making a handler function in order to update the selectedTabId through the  setSelectedTab.

src/App.tsx
```diff
@@ -1,42 +1,37 @@
...
+
+  const selectedTab = tabs.find(tab => tab.id === selectedTabId);
```
kenvastolord, line 22: Well done! is a good idea to use the find method and store the result in a variable.

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,55 @@
...
+
+  return (
+    <div data-cy="TabsComponent">
+      <div className="tabs is-boxed">
+        <ul>
+          {tabs.map(tab => (
+            <li
+              key={tab.id}
+              className={cn({
+                'is-active': selectedTab.id === tab.id,
```
kenvastolord, line 27: do you consider that  use just 'cn' to name the classnames is a good idea? I would suggest to use a descriptive names for your variables.

src/App.tsx
```diff
@@ -1,42 +1,37 @@
...
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
   { id: 'tab-2', title: 'Tab 2', content: 'Some text 2' },
   { id: 'tab-3', title: 'Tab 3', content: 'Some text 3' },
 ];
 
-export const App: React.FC = () => {
+export const App: FC = () => {
+  const [selectedTabId, setSelectedTab] = useState(tabs[0].id);
+
+  const tabSelectHandler = (tab: Tab) => {
```
naziamoff, line 18:
```suggestion
  const selectTab = (tab: Tab) => {
```

src/components/Tabs/Tabs.tsx
naziamoff: @kenvastolord that's ok in that case, classnames is very popular lib and it is clear what we're dealing with by `cn`

naziamoff (approved):
LGTM

PR #1373, opened 2023-01-23
descriptionShouldBeUpdated

grslckr:


src/App.tsx
```diff
@@ -10,33 +11,47 @@ export const tabs = [
...
+              onTabSelected={setSelectedTab}
+            />
           </ul>
         </div>
 
         <div className="block" data-cy="TabContent">
-          Some text 1
+          {selectedTab.content}
         </div>
       </div>
```
CharAnagn, line 53: I believe you should move this whole block to the Tabs component

src/App.tsx
```diff
@@ -10,33 +11,47 @@ export const tabs = [
...
+  useEffect(() => {
+    if (tabs.some(tab => tab.id === selectedTabId)) {
+      return;
+    }
+
+    setSelectedTab({
+      id: 'tab-1', title: 'Tab 1', content: 'Some text 1',
+    });
+  }, [selectedTabId]);
+
```
CharAnagn, line 32: The double useEffect with dependecies that they are so connected and are not side effects, it's a bit confusing and I believe uneccessary for that logic, because either way if your state variable changes the component will re-render and so will do, if through your child component the setter method for your state updates the state.

I suggest you can create a helper function that you will pass as props to the tabs, instead of the setter method directly, that will take the tab as parameter and will update your selectedTabId based on that tab.
 
And try to use the find method to find a tab that matches your selectedTabId so you can use that for your title. And you don't have to worry if your title will change when your selectedTabId changes, because react will do it on it's own, it will re-render and run again from top to bottom the component.

CharAnagn (approved):
I will approve so the mentors maybe can help you a bit better than my comment. But you have the right idea just you need some adjustments.

naziamoff (changes requested):
good review from @CharAnagn, let's implement his points

meforoma (approved):
it's functional as expected.
You could play around with passing tests in spare time.
look for `.spec.ts` files to understand what the test does

PR #1376, opened 2023-01-24
Finished

JzSol:
[DEMO LINK](https://JzSol.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,8 @@
-import React from 'react';
+import { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
+import { Tabs } from './components/Tabs';
 
 export const tabs = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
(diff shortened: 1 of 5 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,57 @@
...
+              >
+                {tab.title}
+              </a>
+            </li>
+          ))}
+        </ul>
+      </div>
+
+      <div className="block" data-cy="TabContent">
+        {tabs.find((tab) => tab.id === id)?.content}
```
kenvastolord, line 53: you already has the tab in the variable validTab, it would be better to use { validTab?.content }

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,57 @@
...
+  onTabSelected: (tab: Tab) => void;
+  selectedTabId: string;
+};
+
+export const Tabs: React.FC<Props> = ({
+  onTabSelected,
+  selectedTabId,
+  tabs,
+}) => {
+  const validTab = tabs.find((tab) => tab.id === selectedTabId) || tabs[0];
```
kenvastolord, line 22: Fantastic idea to use the find method and store the result in a variable.

kenvastolord (commented):
excellent job working on this task.

src/App.tsx
```diff
@@ -10,33 +11,18 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [currentTab, setCurrentTab] = useState(tabs[0]);
+  const findTabs = tabs.find((tab) => tab.id === currentTab.id)?.title;
```
naziamoff, line 15: what is findTabs? What do you store there? Looks unclear

src/components/Tabs/Tabs.tsx
naziamoff, line 22:
```suggestion
  const currentTab = tabs.find((tab) => tab.id === selectedTabId) || tabs[0];
```

src/components/Tabs/Tabs.tsx
naziamoff: agree with @kenvastolord

naziamoff (changes requested):

meforoma (approved):

naziamoff (approved):

PR #1377, opened 2023-01-24
solution

lukacsja:
[DEMO LINK](https://lukacsja.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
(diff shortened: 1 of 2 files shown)


src/App.tsx
```diff
@@ -10,32 +12,26 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [selectedTabId, setSelectedTabId] = useState(tabs[0].id);
```
lefteris-L, line 15: If you saved the whole tab instead of just id, then you wouldn't have to use .find() at all

lefteris-L (approved):
Great job

src/App.tsx
naziamoff: good point from @lefteris-L

naziamoff (approved):

PR #1393, opened 2023-02-01
Tabs navigation was realised

anna-inozemets:
[DEMO LINK](https://anna-inozemets.github.io/react_tabs/)

### DIFF
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -1,1 +1,52 @@
-export const Tabs = () => {};
+import classNames from 'classnames';
+import { Tab } from '../../types/Tab';
+
+type Props = {
+  tabs: Tab[],
+  selectedTabId: string,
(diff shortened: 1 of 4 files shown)


svitjojo (approved):
Brilliant💠

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,52 @@
...
+  };
+
+  return (
+    <div data-cy="TabsComponent">
+      <div className="tabs is-boxed">
+        <ul>
+          {tabs.map(tab => (
+            <li
+              className={classNames(
+                { 'is-active': tab.id === selectedTab.id },
```
Anna-Tkachenko, line 30: you checked is current tab selected several times, here and in `handleSelect` function
will be better to use separate boolean variable, use it here and pass in `handleSelect` function

Anna-Tkachenko (approved):
Good job!

PR #1406, opened 2023-02-06
add solution

Anastasiia-Khudych:
[Demo](https://anastasiia-khudych.github.io/react_tabs/#tab-1)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
+import { Tabs } from './components/Tabs';
 
 export const tabs = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -10,33 +11,19 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [selectedTab, setSelectedTab] = useState(tabs[0]);
+
   return (
     <div className="section">
       <h1 className="title">
-        Selected tab is Tab 1
+        {`Selected tab is ${selectedTab.title}`}
```
Oleksandr-Maly, line 19: you can use destruction for selectedTab.title and selectedTab.id

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,59 @@
...
+};
+
+export const Tabs: React.FC<Props> = (
+  {
+    tabs,
+    selectedTabId,
+    onTabSelected,
+  },
+) => {
+  const selectedTab = tabs.find(tab => tab.id === selectedTabId) || tabs[0];
```
Oleksandr-Maly, line 18: i think you already have selectedTab, it kept in the state of the App

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,59 @@
...
+
+  const handleTabClick = (tab: Tab) => {
+    if (selectedTabId !== tab.id) {
+      onTabSelected(tab);
```
Oleksandr-Maly, line 22: think, you can use this: 
`
const handleOnClick = () => {
    onTabSelected(tab);
  };
`

Oleksandr-Maly (approved):
Good job!

src/components/Tabs/Tabs.tsx
denys-cheporniuk: yeah, but it's better to pass id to `Tabs` instead of tab object
Screenshot 2023-02-07 at 14 37 31

src/components/Tabs/Tabs.tsx
denys-cheporniuk: we can leave this `if` we don't want to make additional actions if the user clicks on the same tab

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,59 @@
...
+  };
+
+  return (
+    <div data-cy="TabsComponent">
+      <div className="tabs is-boxed">
+        <ul>
+          {tabs.map(tab => {
+            const { id, title } = tab;
+
+            return (
```
denys-cheporniuk, line 33:
```suggestion
        <ul>
          {tabs.map(({ id, title })=> (
```

denys-cheporniuk (approved):

PR #1407, opened 2023-02-07
solution

Oleksandr-Maly:
[DEMO LINK](https://oleksandr-maly.github.io/react_tabs/)


Bogdan-Kutasevich (changes requested):
Hi, Good job!) But some is need to fix.
You don't need to add additional TabInfo component, all logic must be in Tabs. It`s maybe reason why you didn't pass tests. If you don't agree with it change status for mentor review)

denys-cheporniuk (approved):

PR #1417, opened 2023-02-08
solution

imariash86:
[DEMO LINK](https://imariash86.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import React, { useState } from 'react';
+import { Tabs } from './components/Tabs';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -10,33 +11,20 @@ export const tabs = [
...
-
-            <li data-cy="Tab">
-              <a href="#tab-3" data-cy="TabLink">Tab 3</a>
-            </li>
-          </ul>
-        </div>
+      <Tabs
+        tabs={tabs}
+        selectedTabId={selectedTab.id}
+        onTabSelected={tab => setSelectedTab(tab)}
```
OlehSharha, line 25: i think u can create separate function for this

denys-cheporniuk: agree, here should be separate function

denys-cheporniuk (approved):

PR #1426, opened 2023-02-08
solution

victorionov:
https://victorionov.github.io/react_tabs/

### DIFF
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -1,1 +1,54 @@
-export const Tabs = () => {};
+import React from 'react';
+import classNames from 'classnames';
+
+export interface Tab {
+  id: string,
+  title: string,
+  content: string,
+}
+
+interface TabsPropsType {
+  tabs: Tab[],
+  selectedTabId: string,
+  onTabSelected: (tab: Tab) => void,
+}
+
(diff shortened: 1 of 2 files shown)


svitjojo (approved):
Good job 👍

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,54 @@
...
+
+    return (
+      <div data-cy="TabsComponent">
+        <div className="tabs is-boxed">
+          <ul>
+            {tabs.map(tab => (
+              <li
+                key={tab.id}
+                className={classNames({
+                  'is-active': selectedTab.id === tab.id,
```
Anna-Tkachenko, line 34: you checked is current tab selected several times
will be better to use separate boolean variable and use it here and pass it in `setSelectedTab` function

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,54 @@
-export const Tabs = () => {};
+import React from 'react';
+import classNames from 'classnames';
+
+export interface Tab {
```
Anna-Tkachenko, line 4: create separate file for `Tab` type

Anna-Tkachenko (approved):

PR #1431, opened 2023-02-09
react tabs was done

Tanya-Grebenkina:
https://Tanya-Grebenkina.github.io/react_tabs/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,42 +1,34 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
+import { Tabs } from './components/Tabs';
+import { Tab } from './types/Tab';
 
-export const tabs = [
+export const tabsFromServer = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
   { id: 'tab-2', title: 'Tab 2', content: 'Some text 2' },
   { id: 'tab-3', title: 'Tab 3', content: 'Some text 3' },
 ];
 
 export const App: React.FC = () => {
+  const [selectedTab, setSelectedTab] = useState(tabsFromServer[0]);
+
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,42 +1,34 @@
...
-          </ul>
-        </div>
-
-        <div className="block" data-cy="TabContent">
-          Some text 1
-        </div>
-      </div>
+      <Tabs
+        tabs={tabsFromServer}
+        selectedTabId={selectedTab.id}
```
Anastasiia-Svintsova, line 29: Look, you have the whole tab in a state, but pass only selectedTab.id to the Tabs component. And after that, you find a selected tab there. Why don't you pass the whole tab at once

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,55 @@
...
+  selectedTabId: string;
+};
+
+export const Tabs: React.FC<Props> = ({
+  tabs,
+  onTabSelected,
+  selectedTabId,
+}) => {
+  const isSelectedTab = tabs.find(tab => selectedTabId === tab.id)
+  || tabs[0];
```
Anastasiia-Svintsova, line 16: So when you'll fix the comment above, you'll have the possibility to destructure the selected tab here

Anastasiia-Svintsova (changes requested):

src/App.tsx
Tanya-Grebenkina: I didn't fully understand exactly what you meant, but if I understood correctly.... it gave me an idea of how to write it differently and I revised the logic in general😅

Tanya-Grebenkina: @Anastasiia-Svintsova  I have already understood what you mean....I passed the whole object(not just "id")... and it works....but somehow 4 tests failed...maybe the problem in the tests themselves? maybe they expect to accept "id" instead of "object"?

nazarmatsevych (approved):

PR #1441, opened 2023-02-12
solution

Kravchenko71213:
https://Kravchenko71213.github.io/react_tabs/

### DIFF
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -1,1 +1,57 @@
-export const Tabs = () => {};
+import React from 'react';
+import classNames from 'classnames';
+import { Tab } from '../../type/Tab';
+
+type Props = {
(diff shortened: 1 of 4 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,57 @@
...
+  tabs,
+  selectedTabId,
+  onTabSelected,
+}) => {
+  const selectedTab = tabs.find(tab => tab.id === selectedTabId) || tabs[0];
+
+  const handleClickTab = (tab: Tab) => {
+    if (tab.id === selectedTabId) {
+      return;
+    }
```
Anna-Tkachenko, line 21: will be better to process some functionality according to some check, and just don't write anything if this check returns `false` value

Anna-Tkachenko (approved):

PR #1457, opened 2023-02-19
React_Tabs was done

illia-kots:
[DEMO LINK](https://illia-kots.github.io/react_tabs/)
I don't understand why, but some tests failed. Although the code works the same way as in the example version.

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
+
+import { Tabs } from './components/Tabs';
+import { Tab } from './components/types/Tab';
+
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -10,31 +14,29 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [selectedTabId, setselectedTabId] = useState(tabs[0]);
+
+  const tabSelection = (tab: Tab) => {
+    setselectedTabId(tab);
+  };
```
Anastasiia-Svintsova, line 21: You should check if tab !== selectedTabId
image

src/App.tsx
```diff
@@ -10,31 +14,29 @@ export const tabs = [
...
+          <Tabs
+            tabs={tabs}
+            selectedTabId={selectedTabId.id}
+            onTabSelected={tabSelection}
+          />
         </div>
 
         <div className="block" data-cy="TabContent">
-          Some text 1
+          {selectedTabId.content}
```
Anastasiia-Svintsova, line 39: The content should be inside the Tabs component. You should find a tab content according to the tab.id and show found tab or tabs[0] to pass this test
image

src/App.tsx
```diff
@@ -10,31 +14,29 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [selectedTabId, setselectedTabId] = useState(tabs[0]);
```
Anastasiia-Svintsova, line 17: It's not id here, it's a tab
```suggestion
  const [selectedTab, setselectedTab] = useState(tabs[0]);
```

Anastasiia-Svintsova (changes requested):

OleziO (approved):

src/App.tsx
illia-kots: These tests are still failing!

Anastasiia-Svintsova (approved):
Your work has been approved already, so you don't need to request a review again. But my comments weren’t fixed :smiling_face_with_tear:

illia-kots:
I know that my request has been approved🙂, but I wanted to solve this problem with my tests.

PR #1467, opened 2023-03-03
React_tabs

07Spirit:
[DEMO LINK](https://07Spirit.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -13,7 +13,7 @@ export const tabs = [
 export const App: React.FC = () => {
   const [selectedTabId, selectedTabIdEdit] = useState('tab-1');
   const selectedForm = tabs.filter((tab) => tab.id === selectedTabId);
-  const [content, title] = selectedForm;
+  const [{ content, title }] = selectedForm;
 
   return (
     <div className="section">
@@ -22,14 +22,11 @@ export const App: React.FC = () => {
       <div data-cy="TabsComponent">
         <div className="tabs is-boxed">
           <ul>
-            {tabs.map((tab) => (
-              <Tabs
-                key={tab.id}
-                tab={tab}
-                selectId={selectedTabId}
-                onTabSelected={selectedTabIdEdit}
-              />
-            ))}
+            <Tabs
+              tabs={tabs}
+              selectedTabId={selectedTabId}
+              onTabSelected={selectedTabIdEdit}
+            />
           </ul>
         </div>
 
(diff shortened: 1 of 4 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,38 @@
...
+
+export const Tabs = ({
+  tabs,
+  selectedTabId,
+  onTabSelected,
+}: {
+  tabs: TabsType[];
+  selectedTabId: string;
+  onTabSelected: React.Dispatch<React.SetStateAction<string>>;
+}) => {
```
Anastasiia-Svintsova, line 17: Move it to props type, please 
```suggestion
}: Props) => {
```

or you can use React.FC
```
export const Tabs: React.FC<Props> = ...

File: src/components/Tabs/Tabs.tsx [hunk 2]
```diff
@@ -1 +1,38 @@
...
+        <li
+          key={tab.id}
+          data-cy="Tab"
+          className={classNames({ 'is-active': tab.id === selectedTabId })}
+        >
+          <a
+            href={`#${tab.id}`}
+            data-cy="TabLink"
+            id={tab.id}
+            onClick={(event) => onTabSelected(event.currentTarget.id)}
```
Anastasiia-Svintsova, line 30: First of all, move it to a separate function. And follow the requirements 
image

Also, you don't really need an event here 🤔  You already have a `tab.id`

src/App.tsx
```diff
@@ -10,31 +11,27 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [selectedTabId, selectedTabIdEdit] = useState('tab-1');
```
Anastasiia-Svintsova, line 14:
```suggestion
  const [selectedTabId, selectedTabIdEdit] = useState(tabs[0].id);
```

src/App.tsx
```diff
@@ -10,31 +11,27 @@ export const tabs = [
...
+  const selectedForm = tabs.filter((tab) => tab.id === selectedTabId);
```
Anastasiia-Svintsova, line 15: seems that currentTab would fit better. But it's up to you. And better to use the find method here. You need only one tab
```suggestion
  const currentTab = tabs.find((tab) => tab.id === selectedTabId) || tabs[0];
```

Anastasiia-Svintsova (changes requested):

qoa11a (approved):

PR #1468, opened 2023-03-03
done

AntonChetverykMA:
[DEMO LINK](https://AntonChetverykMA.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
+import { Tabs } from './components/Tabs';
 
 export const tabs = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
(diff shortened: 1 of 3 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,44 @@
...
+
+  return (
+    <div data-cy="TabsComponent">
+      <div className="tabs is-boxed">
+        <ul>
+          {tabs.map(tab => {
+            return (
+              <li
+                // eslint-disable-next-line max-len
+                className={classNames({ 'is-active': tab.id === selectedTabId })}
```
OleziO, line 22: Divide this line by a few lines. Try to avoid linter comments, it's bad practice.

OleziO (changes requested):

src/App.tsx
```diff
@@ -10,33 +11,29 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [selectedTabId, setSelectedTabId] = useState('tab-1' || tabs[0]);
+  const selectedTab: string = tabs.find(tab => tab.id === selectedTabId)?.title
+                              || tabs[0].title;
```
TarasHoliuk, line 16: bad formatting

```suggestion
  const selectedTab: string = tabs
    .find(tab => tab.id === selectedTabId)?.title || tabs[0].title;
```

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,45 @@
...
+  const { tabs, selectedTabId, onTabSelected } = props;
+  const content = tabs.find(tab => tab.id === selectedTabId)?.content;
+
+  return (
+    <div data-cy="TabsComponent">
+      <div className="tabs is-boxed">
+        <ul>
+          {tabs.map(tab => {
+            return (
+              <li
```
TarasHoliuk, line 20:
```suggestion
          {tabs.map(tab => (
            <li
```
you can do it in this way

TarasHoliuk (changes requested):

TarasHoliuk (changes requested):
I didn't get why you deleted `package.json`, doesn't seem the common solution for any issue :)

If there was some special reason to delete it, please add some explanation (maybe lin to Slack thread, etc.)

AlisaVlasova (approved):

PR #1472, opened 2023-03-08
React tabs solution

kirill-vypirovskyi:
[DEMO LINK](https://kirill-vypirovskyi.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,42 +1,44 @@
-import React from 'react';
+import { FC, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
+import { Tabs } from './components/Tabs';
+import { Tab } from './types/tab';
 
-export const tabs = [
+export const tabs: Tab[] = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
   { id: 'tab-2', title: 'Tab 2', content: 'Some text 2' },
   { id: 'tab-3', title: 'Tab 3', content: 'Some text 3' },
 ];
 
-export const App: React.FC = () => {
+export const App: FC = () => {
+  const [selectedTabId, changeTab] = useState(tabs[0].id);
+
+  const toggleTab = (tab: Tab) => {
+    changeTab(tab.id);
+  };
+
+  const getActiveTabTitle = () => {
+    const currentTab = tabs.find((tab) => tab.id === selectedTabId);
+
+    if (currentTab) {
+      return currentTab.title;
+    }
+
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,42 +1,44 @@
...
+  };
+
+  const getActiveTabTitle = () => {
+    const currentTab = tabs.find((tab) => tab.id === selectedTabId);
+
+    if (currentTab) {
+      return currentTab.title;
+    }
+
+    return tabs[0].title;
```
kuznetsova-anastasiia, line 28: how about use conditional operator for return? (on your taste)

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,58 @@
...
+                  'is-active': id === choosedTab.id,
+                })}
+                data-cy="Tab"
+                key={id}
+              >
+                <a
+                  href={`#${id}`}
+                  data-cy="TabLink"
+                  onClick={() => {
+                    if (id !== selectedTabId) {
```
kuznetsova-anastasiia, line 40: maybe `choosedTab.id`?  because `selectedTabId` can be invalid that means that our selected tab is the first tab, and what if that the first tab was already selected?
```suggestion
                    if (id !== choosedTab.id) {
```

kuznetsova-anastasiia (approved):
GJ

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,58 @@
...
+                      onTabSelected(tab);
+                    }
+                  }}
```
illiarerun, line 43: You're recreating this handler for every element of the list, move it outside of jsx

src/components/Tabs/Tabs.tsx
illiarerun: @kuznetsova-anastasiia then it won't be reselected

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,58 @@
...
+          {tabs.map((tab) => {
+            const {
+              id,
+              title,
+            } = tab;
+
+            return (
+              <li
+                className={classNames({
+                  'is-active': id === choosedTab.id,
```
illiarerun, line 31: why not `selectedTabId`?

illiarerun (changes requested):

src/components/Tabs/Tabs.tsx
kirill-vypirovskyi: idk, just the first word than i remembered

illiarerun (approved):

PR #1477, opened 2023-03-11
React_tabs

SerhiiVovchenko:
[DEMO LINK](https://SerhiiVovchenko.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,9 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
+import { Tabs } from './components/Tabs';
+import { Tab } from './types/Tab';
 
 export const tabs = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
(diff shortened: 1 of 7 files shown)


src/App.tsx
```diff
@@ -10,33 +12,24 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [selectedTabId, setSelectedTabId] = useState(tabs[0].id);
+  const currentСomponent = tabs.find(({ id }) => id === selectedTabId);
```
polosanya, line 16:
```suggestion
  const selectedTab = tabs.find(({ id }) => id === selectedTabId);
```
I think this name would be more accurate. 
Also, what if find methods finds nothing and return undefined? You should set the first tab in that case.
image

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,45 @@
...
+import React from 'react';
+import classNames from 'classnames';
+import { TabsType } from '../../types/Tabs';
+
+export const Tabs: React.FC<TabsType> = ({
+  tabs,
+  selectedTabId,
+  onTabSelected,
+}) => {
+  const currentСomponent = tabs.filter(tab => tab.id === selectedTabId);
```
polosanya, line 10: Same here. Also, why do you use a filter here instead of find?
Only one tab can be selected at a time

polosanya (changes requested):
Let's try to pass all tests)

agniya-i (approved):

PR #1481, opened 2023-03-13
done

l4st1m0za:
[DEMO LINK](https://l4st1m0za.github.io/react_tabs/)

### DIFF
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -1,1 +1,57 @@
-export const Tabs = () => {};
+import React from 'react';
+import classNames from 'classnames';
+import { Tab } from '../../types/Tab';
+
+type Props = {
+  tabs: Tab[],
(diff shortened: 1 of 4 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,57 @@
...
+                data-cy="Tab"
+              >
+
+                <a
+                  href={`#${id}`}
+                  data-cy="TabLink"
+                  onClick={() => {
+                    if (selectedTabId !== id) {
+                      onTabSelected({ id, title, content });
+                    }
```
Volodymyr-Melnyk-Fron10der, line 38: Move this into separated function, and you actually don't need to check this condition inside handleTab if you do it here

Volodymyr-Melnyk-Fron10der (changes requested):
Almost there! Do not forget about DEMO_LINK in description

PR #1484, opened 2023-03-14
Solution

katerinalex:
[DEMO LINK](https://katerinalex.github.io/react_tabs/)


src/App.tsx
```diff
@@ -40,3 +41,36 @@ export const App: React.FC = () => {
     </div>
   );
 };
+
+type State = {
+  selectedTabId: string,
+};
+
+export class App extends React.Component<{}, State> {
```
anstsot, line 49: it would be better to use functional components in the future.

anstsot (approved):
Great job, approved!

PR #1487, opened 2023-03-15
solution

vlad-trundenko:
[DEMO LINK](https://vlad-trundenko.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -21,13 +21,9 @@ export class App extends Component<{}, State> {
   };
 
   onTabSelected = (tab: Tab) => {
-    const { selectedTabId } = this.state;
-
-    if (selectedTabId !== tab.id) {
-      this.setState({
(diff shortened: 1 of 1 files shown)


src/App.tsx
```diff
@@ -1,42 +1,54 @@
...
+
+export class App extends Component<{}, State> {
+  state = {
+    selectedTabId: tabs[0].id,
+  };
+
+  onTabSelected = (tab: Tab) => {
+    const { selectedTabId } = this.state;
+
+    if (selectedTabId !== tab.id) {
```
Volodymyr-Melnyk-Fron10der, line 26: you are checking it in your handleTab  funcction, so there is no need to do it again here

Volodymyr-Melnyk-Fron10der (changes requested):
Well done, just one small fix plz

witflash (approved):
Well done 👍

PR #1493, opened 2023-03-17
Solution

lilia-obushenko:
[DEMO LINK](https://lilia-obushenko.github.io/react_tabs/)

### DIFF
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -1,1 +1,52 @@
-export const Tabs = () => {};
+import { FC } from 'react';
+import cn from 'classnames';
+import { Tab } from '../../types/Tab';
+
+interface Props {
+  tabs: Tab[]
+  selectedTabId: string,
+  onTabSelected: (tab: Tab) => void
+}
+
+export const Tabs: FC<Props> = (props) => {
+  const { tabs, selectedTabId, onTabSelected } = props;
+  const selectedTab = tabs.find(tab => tab.id === selectedTabId) || tabs[0];
+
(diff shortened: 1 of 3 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,52 @@
...
+      <div className="tabs is-boxed">
+        <ul>
+          {tabs.map(tab => {
+            const { id, title } = tab;
+
+            return (
+              <li
+                key={id}
+                className={cn({
+                  'is-active': id === selectedTab.id,
```
rashyd-hasratov, line 26: You have comparisons like `id === selectedTab.id` twice here, so you just could create a corresponding value like `const isSelected = id === selectedTab.id`

rashyd-hasratov (changes requested):
Great! Just a little fix

rashyd-hasratov (approved):

src/components/Tabs/Tabs.tsx
OleksiiBlezniuk: Good point! Actually it's pretty similar check

```
const isSelected = id === selectedTab.id;

isSelected // id === selectedTab.id
!isSelected // id !== selectedTab.id
```

OleksiiBlezniuk (changes requested):

src/components/Tabs/Tabs.tsx
lilia-obushenko: My fault. Fixed)

OleksiiBlezniuk (approved):
GJ!

PR #1495, opened 2023-03-20
● Solution

d-hubych:
- [DEMO LINK](https://d-hubych.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
+import { Tab } from './types/Tab';
+import { Tabs } from './components/Tabs';
+
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -10,33 +13,23 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [selectedTabId, setSelectedTabId] = useState(tabs[0].id);
+
+  const onTabSelected = (selectedTub: Tab) => {
```
etojeDenys, line 18: we can pass only id instead of whole Tab

etojeDenys (changes requested):
almost done

src/App.tsx
d-hubych: tests require an object

src/App.tsx
```diff
@@ -10,33 +13,23 @@ export const tabs = [
...
+
+  const handleSelectTab = ({ id }: Tab) => {
+    setSelectedTabId(id);
+  };
+
   return (
     <div className="section">
       <h1 className="title">
-        Selected tab is Tab 1
+        { `Selected tab is Tab ${selectedTabId.slice(-1)}` }
```
DarkMistyRoom, line 25: you can use `title `here instead of slicing

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,50 @@
...
+              key={tab.id}
+            >
+              <a
+                href={`#${tab.id}`}
+                data-cy="TabLink"
+                onClick={() => {
+                  if (tab.id !== selectedTabId) {
+                    onTabSelected(tab);
+                  }
+                }}
```
DarkMistyRoom, line 36: move the handler to a separate function

DarkMistyRoom (approved):

PR #1497, opened 2023-03-22
Solution by DK

DimaK123:
[DEMO LINK](https://Dimak123.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
+import { Tabs } from './components/Tabs';
 
 export const tabs = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -10,33 +12,29 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [selectedTabId, setSelectedTabId] = useState('tab-1');
+
+  const currentTab = tabs.find(tab => tab.id === selectedTabId) || tabs[0];
+
+  const onTabSelected = (tab: Tab) => {
```
polosanya, line 19:
```suggestion
  const handleTabSelected = (tab: Tab) => {
```
https://javascript.plainenglish.io/handy-naming-conventions-for-event-handler-functions-props-in-react-fc1cbb791364

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,47 @@
...
+        <div className="tabs is-boxed">
+          <ul>
+            {tabs.map(tab => {
+              const { id, title } = tab;
+
+              return (
+                <li
+                  key={id}
+                  data-cy="Tab"
+                  className={classNames({ 'is-active': selectedTabId === id })}
```
polosanya, line 30: Compare it with the currentTab.id

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,47 @@
...
+                >
+                  <a href={`#${id}`} data-cy="TabLink" onClick={() => currentTab !== tab && onTabSelected(tab)}>
```
polosanya, line 32:
```suggestion
                  <a 
                    href={`#${id}`} 
                    data-cy="TabLink" 
                    onClick={() => currentTab !== tab && onTabSelected(tab)}
                  >
```
Also, don't compare objects like that
image

src/App.tsx
```diff
@@ -10,33 +12,29 @@ export const tabs = [
...
+    if (tabs.find(someTab => someTab.id === tab.id)) {
+      setSelectedTabId(tab.id);
+    } else {
+      setSelectedTabId(tabs[0].id);
```
polosanya, line 23: You already check it when finding a tab, so there is actually no need to double check

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,47 @@
...
+  onTabSelected: (tab: Tab) => void,
+};
+
+export const Tabs:React.FC<Props> = React.memo(
+  ({
+    tabs,
+    selectedTabId,
+    onTabSelected,
+  }) => {
+    const currentTab = tabs.find(tab => tab.id === selectedTabId) || tabs[0];
```
polosanya, line 17: You do the same here and in App. 
You can create a helper function in order not to duplicate yourself

polosanya (changes requested):
Let's improve your code a bit)

src/components/Tabs/Tabs.tsx
DimaK123: Thanks for the advice. I hope that I've done everything correctly, but unfortunately, the test still failed. I see the error message in the terminal.
Why onTabFinder is not a function ?


изображение

TarasHoliuk (approved):
LGTM :+1:

agniya-i (approved):

PR #1550, opened 2023-04-06
commit message

diamat1997:
[DEMO LINK](https://diamat1997.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -10,33 +11,19 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [selectedTab, setSelectedTab] = useState(tabs[0]);
```
DarkMistyRoom, line 14: do you need to store the whole tab. id should be enough

DarkMistyRoom (approved):
Good job!

PR #1554, opened 2023-04-07
Solution

OlhaKissa:
[DEMO LINK](https://olhakissa.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,9 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -10,33 +12,23 @@ export const tabs = [
...
+
+  const handleTabSelect = (tab: Tab) => {
+    setSelectedTabId(tab.id);
+  };
+
   return (
     <div className="section">
       <h1 className="title">
-        Selected tab is Tab 1
+        {`Selected tab is ${tabs.find(tab => selectedTabId === tab.id)?.title}`}
```
anstsot, line 24: it would be better to create a separate variable `activeTab` and use it here
```suggestion
        {`Selected tab is ${activeTab.title}`}
```

anstsot (approved):
Good job, approving

PR #1556, opened 2023-04-07
Tabs Solutions

GUSILLUS:


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,58 @@
...
+};
+
+export const Tabs: React.FC<Props> = (
+  {
+    tabs,
+    selectedTabId,
+    onTabSelected,
+  },
+) => {
+  const currentTab = tabs.find(tab => tab.id === selectedTabId) || tabs[0];
```
etojeDenys, line 22: In case 'tabs' array is empty, trying to access 'tabs[0]' will return 'undefined', which will cause issues when accessing 'content'. Use optional chaining and nullish coalescing to handle this case.

etojeDenys (changes requested):
your forgot to add demo link to your PR

etojeDenys (changes requested):
add demo link to your PR

src/App.tsx
```diff
@@ -1,42 +1,43 @@
...
+    setSelectedTabId(tab.id);
+  };
+
   return (
     <div className="section">
       <h1 className="title">
-        Selected tab is Tab 1
+        Selected tab is
+        {' '}
+        {title}
```
OleziO, line 33: Replace with template string.

[More info.](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals)

OleziO (changes requested):

PR #1561, opened 2023-04-10
error checking

FallenMAD:
[DEMO LINK](https://FallenMAD.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
+import { Tabs } from './components/Tabs';
 
 export const tabs = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
@@ -10,33 +11,31 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [selectedTabId, setSelectedTabId] = useState(tabs[0]);
+
+  // const hanldeTabSelected = (tab: Tab) => {
+  //   if (tab.id !== selectedTabId.id) {
+  //     setSelectedTabId(tab.id);
+  //   }
+  // };
+
   return (
     <div className="section">
       <h1 className="title">
-        Selected tab is Tab 1
+        Selected tab is
+        {' '}
+        {selectedTabId.title}
       </h1>
 
-      <div data-cy="TabsComponent">
-        <div className="tabs is-boxed">
-          <ul>
-            <li className="is-active" data-cy="Tab">
-              <a href="#tab-1" data-cy="TabLink">Tab 1</a>
-            </li>
+      <Tabs
+        tabs={tabs}
+        selectedTab={selectedTabId}
+        onTabSelected={setSelectedTabId}
+      />
 
-            <li data-cy="Tab">
-              <a href="#tab-2" data-cy="TabLink">Tab 2</a>
-            </li>
-
-            <li data-cy="Tab">
-              <a href="#tab-3" data-cy="TabLink">Tab 3</a>
-            </li>
-          </ul>
-        </div>
-
-        <div className="block" data-cy="TabContent">
-          Some text 1
-        </div>
-      </div>
+      {/* <div className="block" data-cy="TabContent">
+        {selectedTabId.content}
+      </div> */}
     </div>
   );
 };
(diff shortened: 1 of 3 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,57 @@
...
+}) => {
+  const hanldeTabSelected = (
+    tab: Tab,
+    event: React.SyntheticEvent<Element, Event>,
+  ) => {
+    event.preventDefault();
+
+    if (tab.id === selectedTab.id) {
+      onTabSelected(selectedTab);
+    }
```
Anastasiia-Svintsova, line 23: Don't call this function if id is the same 
image
```suggestion
    if (tab.id !== selectedTab.id) {
      return onTabSelected(tab);
    }
```

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,57 @@
...
+              key={tab.id}
+              className={classNames({
+                'is-active': selectedTab === tab,
+              })}
+              data-cy="Tab"
+            >
+              <a
+                href={`#${tab.id}`}
+                data-cy="TabLink"
+                onClick={(event) => hanldeTabSelected(tab, event)}
```
Anastasiia-Svintsova, line 43: Don't need to pass the event here
```suggestion
                onClick={() => hanldeTabSelected(tab)}
```

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,57 @@
...
+export const Tabs: React.FC<Props> = ({
+  tabs,
+  selectedTab,
+  onTabSelected,
+}) => {
+  const hanldeTabSelected = (
+    tab: Tab,
+    event: React.SyntheticEvent<Element, Event>,
+  ) => {
+    event.preventDefault();
```
Anastasiia-Svintsova, line 19: The link doesn't work because of this action
```suggestion
```

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,57 @@
...
+              >
+                {tab.title}
+              </a>
+            </li>
+          ))}
+        </ul>
+      </div>
+
+      <div className="block" data-cy="TabContent">
+        {selectedTab.content}
```
Anastasiia-Svintsova, line 53: Use destructuring for the selectedTab

src/App.tsx
```diff
@@ -10,33 +11,31 @@ export const tabs = [
...
+        onTabSelected={setSelectedTabId}
+      />
 
-        <div className="block" data-cy="TabContent">
-          Some text 1
-        </div>
-      </div>
+      {/* <div className="block" data-cy="TabContent">
+        {selectedTabId.content}
+      </div> */}
```
Anastasiia-Svintsova, line 38: Remove comments

Anastasiia-Svintsova (changes requested):
Good job! Keep going. Will continue the conversation in the chat

src/App.tsx
```diff
@@ -10,33 +11,21 @@ export const tabs = [
...
 export const App: React.FC = () => {
+  const [selectedTabId, setSelectedTabId] = useState(tabs[0]);
+
   return (
     <div className="section">
       <h1 className="title">
-        Selected tab is Tab 1
+        Selected tab is
+        {' '}
+        {selectedTabId.title}
```
witflash, line 21: You can use a string interpolation instead of three separate lines. It should looks better

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,54 @@
...
+  onTabSelected: (tab: Tab) => void,
+};
+
+export const Tabs: React.FC<Props> = ({
+  tabs,
+  selectedTab,
+  onTabSelected,
+}) => {
+  const hanldeTabSelected = (tab: Tab) => {
+    if (tab.id !== selectedTab) {
```
witflash, line 16: So here is a problem. Your `selectedTab` is whole tab object. But you compare id and an object. It isn't work correct

witflash (changes requested):
Good job.
Pay attention to my comments, please

src/App.tsx
```diff
@@ -10,33 +11,19 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [selectedTab, setSelectedTabId] = useState(tabs[0]);
```
Anastasiia-Svintsova, line 14: So, as I said it's not an id
```suggestion
  const [selectedTab, setSelectedTab] = useState(tabs[0]);
```

Anastasiia-Svintsova (approved):

src/App.tsx
FallenMAD: sorry my mistake

PR #1562, opened 2023-04-10
react-tabs1

fenderukr:
(https://fenderukr.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
+import { Tabs } from './components/Tabs';
 
 export const tabs = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
@@ -10,32 +11,22 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [selectedTabId, setSelectedTabId] = useState(tabs[0]);
+
   return (
     <div className="section">
       <h1 className="title">
-        Selected tab is Tab 1
+        Selected tab is
+        {' '}
+        {selectedTabId.title}
       </h1>
 
       <div data-cy="TabsComponent">
-        <div className="tabs is-boxed">
-          <ul>
-            <li className="is-active" data-cy="Tab">
-              <a href="#tab-1" data-cy="TabLink">Tab 1</a>
-            </li>
-
-            <li data-cy="Tab">
-              <a href="#tab-2" data-cy="TabLink">Tab 2</a>
-            </li>
-
-            <li data-cy="Tab">
-              <a href="#tab-3" data-cy="TabLink">Tab 3</a>
-            </li>
-          </ul>
-        </div>
-
-        <div className="block" data-cy="TabContent">
-          Some text 1
-        </div>
+        <Tabs
+          tabs={tabs}
+          selectedTabId={selectedTabId}
+          onTabSelected={setSelectedTabId}
+        />
       </div>
     </div>
   );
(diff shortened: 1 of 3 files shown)


fenderukr:
I had issues with running tests for this task. I consulted with mentors in a chat and they confirmed that I can create a pull request with the errors.

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,51 @@
-export const Tabs = () => {};
+import { SyntheticEvent } from 'react';
+import { Tab, TabsProps } from '../../types/tab';
+
+export const Tabs = ({ tabs, selectedTabId, onTabSelected }: TabsProps) => {
+  const onTabClickHandle = (event: SyntheticEvent, tab: Tab) => {
```
polosanya, line 5: Consider renaming `onTabClickHandle` to `handleTabClick` for better readability and following the naming conventions for event handlers. 📚
https://javascript.plainenglish.io/handy-naming-conventions-for-event-handler-functions-props-in-react-fc1cbb791364

src/App.tsx
```diff
@@ -10,32 +11,22 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [selectedTabId, setSelectedTabId] = useState(tabs[0]);
```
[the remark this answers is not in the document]
fenderukr: i know this and i tried to do this task using the id first. I realized it would be much more logical to store the whole tab there, not only the id as I need access to the title in the app component

implementing this via tab.id in the state would lead to more troubles rather than to the logical solution

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,51 @@
-export const Tabs = () => {};
+import { SyntheticEvent } from 'react';
+import { Tab, TabsProps } from '../../types/tab';
```
vlad-demchuk, line 2: not fixed
image

src/App.tsx
```diff
@@ -10,32 +11,21 @@ export const tabs = [
...
+
   return (
     <div className="section">
       <h1 className="title">
-        Selected tab is Tab 1
+        Selected tab is
+        { ` ${selectedTabId.title}` }
```
vlad-demchuk, line 20: use the mentor's example, you could just use interpolation for the whole titile
image

src/App.tsx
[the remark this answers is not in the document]
vlad-demchuk: then just rename state name

vlad-demchuk (changes requested):

src/components/Tabs/Tabs.tsx
fenderukr: I don't get this problem any longer
What should I fix here?

fenderukr: image
these are the only 2 tests that fail

fenderukr: 295416aece5f61e4b8e7b05a04ddcfd3b5cbcec6

sTorba24: add types/tabs file to github

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,51 @@
...
+    onTabSelected(tab);
+  };
+
+  return (
+    <div data-cy="TabsComponent">
+      <div className="tabs is-boxed">
+        <ul>
+          {tabs.map((tab: Tab) => (
+            <li
+              className={tab.id === selectedTab.id ? 'is-active' : ''}
```
sTorba24, line 27: you can use classNames library 
https://www.npmjs.com/package/classnames

sTorba24 (changes requested):
GJ :) I left you little comments

loralevitska (approved):
👍

PR #1570, opened 2023-04-17
Ready

sergik94:
https://sergik94.github.io/react_tabs/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -12,11 +12,11 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
-  const [selectedTabsId, setTabsId] = useState('tab-1');
+  const [selectedTabsId, setSelectedTabsId] = useState('tab-1');
   const handleClick = (event: React.MouseEvent<HTMLAnchorElement>) => {
     const targetTab = event.currentTarget as HTMLAnchorElement;
 
-    setTabsId(targetTab.hash.slice(1));
+    setSelectedTabsId(targetTab.hash.slice(1));
   };
 
   const selectedTab = tabs.find(tab => tab.id === selectedTabsId);
@@ -32,7 +32,6 @@ export const App: React.FC = () => {
           tabs={tabs}
           selectedTabsId={selectedTabsId}
           onTabSelected={handleClick}
-          selectedTab={selectedTab}
         />
       </div>
     </div>
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -10,14 +10,12 @@ type Props = {
   tabs: TabsType[],
   selectedTabsId: string,
   onTabSelected: React.MouseEventHandler<HTMLAnchorElement>,
-  selectedTab: TabsType | undefined,
 };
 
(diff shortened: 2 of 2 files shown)


src/App.tsx
```diff
@@ -10,32 +12,28 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [selectedTabsId, setTabsId] = useState('tab-1');
```
etojeDenys, line 15: Consider renaming the state update function `setTabsId` to `setSelectedTabId` for better clarity and consistency with the `selectedTabId` state variable.

src/App.tsx
```diff
@@ -10,32 +12,28 @@ export const tabs = [
...
-            </li>
-          </ul>
-        </div>
-
-        <div className="block" data-cy="TabContent">
-          Some text 1
-        </div>
+        <Tabs
+          tabs={tabs}
+          selectedTabsId={selectedTabsId}
```
etojeDenys, line 33: The `Tabs` component should receive only necessary props. Remove the `selectedTab` prop, as it is not needed inside the `Tabs` component.

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,47 @@
...
+  id: string;
+  title: string;
+  content: string;
+}
+
+type Props = {
+  tabs: TabsType[],
+  selectedTabsId: string,
+  onTabSelected: React.MouseEventHandler<HTMLAnchorElement>,
+  selectedTab: TabsType | undefined,
```
etojeDenys, line 13: You can remove the `selectedTab` prop because you already have `selectedTabsId`. You can find the selected tab content using `selectedTabsId` in the component.

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,47 @@
...
+};
+
+export const Tabs: React.FC<Props> = React.memo(({
+  tabs,
+  selectedTabsId,
+  onTabSelected,
+  selectedTab,
```
etojeDenys, line 20: Remove the `selectedTab` prop from the destructuring assignment as it is not needed.

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,47 @@
...
+            >
+              {title}
+            </a>
+          </li>
+        ))}
+      </ul>
+    </div>
+
+    <div className="block" data-cy="TabContent">
+      {selectedTab?.content}
```
etojeDenys, line 44: Instead of using `selectedTab?.content`, you can find the selected tab's content using `selectedTabsId`. For example, you can use `tabs.find(tab => tab.id === selectedTabsId)?.content`.

etojeDenys (changes requested):

src/components/Tabs/Tabs.tsx
sergik94: @etojeDenys , I have tabs.find() operation in the App and I didn't want to dublicate this in the Tabs.

VitaliyBondarenko1982 (approved):
Looks good to me

PR #1571, opened 2023-04-18
Solution 53

mary-kucher:
[DEMO LINK](https://mary-kucher.github.io/react_tabs/)

### DIFF
--- a/src/types/Tab.ts
+++ b/src/types/Tab.ts
@@ -1,0 +1,3 @@
+export interface Tab {
+  id: string, title: string, content: string;
+}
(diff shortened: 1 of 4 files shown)


src/types/Tab.ts
```diff
@@ -0,0 +1,3 @@
+export interface Tab {
+  id: string, title: string, content: string;
```
etojeDenys, line 2: Properties in the interface should be placed on separate lines for better readability 📖

etojeDenys (approved):
LGTM

PR #1579, opened 2023-04-22
Develop

zynkd:
[DEMO LINK](https://zynkd.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -10,33 +11,19 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [selectedTabId, setSelectedTab] = useState(tabs[0]);
```
witflash, line 14: It's not really `tabId` but just whole `tab`.
So you can name it as `selectedTab`.

witflash (approved):
Good work!

PR #1620, opened 2023-04-28
solution

AnastassiaH:
[DEMO](https://anastassiah.github.io/react_tabs/)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,53 @@
...
+      onTabSelected(tabs.find(el => el.id === currTabId));
+    }
+  };
+
+  return (
+    <div data-cy="TabsComponent">
+      <div className="tabs is-boxed">
+        <ul>
+          {tabs.map(tab => (
+            <li
```
Volodymyr-Melnyk-Fron10der, line 28: add key

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,53 @@
...
+              className={tab.id === selectedTabId ? 'is-active' : ''}
+              data-cy="Tab"
+            >
+              <a
+                href={`#${tab.id}`}
+                data-cy="TabLink"
+                onClick={(e) => {
+                  if (e.currentTarget) {
+                    handleClick(e.currentTarget);
+                  }
```
Volodymyr-Melnyk-Fron10der, line 38: move into handleClick

Volodymyr-Melnyk-Fron10der (changes requested):

src/App.tsx
```diff
@@ -10,33 +12,25 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [selectedTabId, setSelectedTabId] = useState('tab-1');
+
+  const onTabSelected = (tab: Tab | undefined) => {
```
TarasHoliuk, line 17: It is better to use the tab id as an argument for the onTabSelected function rather than the whole tab object.

src/App.tsx
```diff
@@ -10,33 +12,25 @@ export const tabs = [
...
+    if (tab) {
+      setSelectedTabId(tab.id);
+    }
```
TarasHoliuk, line 20: It is recommended to use a default value instead of checking if the tab object is undefined.

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,50 @@
...
+export const Tabs:React.FC<Props> = ({
+  tabs,
+  selectedTabId,
+  onTabSelected,
+}) => {
+  const handleClick = (target: EventTarget & HTMLAnchorElement) => {
+    const currTabId = target.getAttribute('href')?.slice(1);
+
+    if (currTabId !== selectedTabId) {
+      onTabSelected(tabs.find(el => el.id === currTabId));
```
TarasHoliuk, line 19: It's better to use a separate function to find the tab and pass its ID to the onTabSelected callback.

TarasHoliuk: Moreover you use it twice in this component

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,50 @@
...
+          {tabs.map(tab => (
+            <li
+              className={tab.id === selectedTabId ? 'is-active' : ''}
+              data-cy="Tab"
+              key={tab.id}
+            >
+              <a
+                href={`#${tab.id}`}
+                data-cy="TabLink"
+                onClick={(e) => handleClick(e.currentTarget)}
```
TarasHoliuk, line 36: I'd pass `tab.id` to `handleClick` - any way you are mapping them and this value will always be available and correct

TarasHoliuk (approved):
Approved since you've fixed the previous comments, but added a few additional - just FYI

PR #1622, opened 2023-04-30
WorkedVersion

AlexRog01:
Tabs component are the worst tests I have ever seen, they did not give me the opportunity to properly implement the solution as I wanted. And in the end, I could not pass the last two tests. No one will help me on the weekends, so I hand it over like this.
[DEMO LINK](https://AlexRog01.github.io/react_tabs/)
image

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,10 @@
-import React from 'react';
+import React, { useState } from 'react';
+
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
+
+import { Tabs } from './components/Tabs';
 
 export const tabs = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -1,42 +1,32 @@
...
-        </div>
-
-        <div className="block" data-cy="TabContent">
-          Some text 1
-        </div>
-      </div>
+      <Tabs
+        tabs={tabs}
+        selectedTabId={selectedTabId}
+        onTabSelected={setSelectedTab}
```
liliia-livandovskaya, line 28: Consider passing a callback function to onTabSelected instead of directly passing setSelectedTab. For example, onTabSelected={(tabId: string) => setSelectedTab(tabId)}

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,48 @@
...
+};
+
+export const Tabs: React.FC<Props> = ({
+  tabs,
+  selectedTabId,
+  onTabSelected,
+}) => {
+  const selectedTab = tabs.find(tab => tab.id === selectedTabId);
+
+  if (!selectedTab) {
```
liliia-livandovskaya, line 16: It's not advised to call the onTabSelected prop directly in the render function, as this might cause unwanted side effects. Consider using useEffect to update the selected tab when the component mounts or the selectedTabId prop changes.

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,48 @@
...
+    onTabSelected('tab-1');
+  }
+
+  const hendlerOnTabSelected = (tabId: string) => {
```
liliia-livandovskaya, line 20: Typo in the function name 'hendlerOnTabSelected', it should be 'handlerOnTabSelected'.

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,48 @@
...
+      onTabSelected(tabId);
+    }
+  };
+
+  return (
+    <div data-cy="TabsComponent">
+      <div className="tabs is-boxed">
+        <ul>
+          {tabs.map(tab => (
+            <li // eslint-disable-line
```
liliia-livandovskaya, line 31: Avoid using 'eslint-disable-line' if possible. In this case, you can get rid of the comment by wrapping the li element with React.Fragment and using the key prop on the React.Fragment.

liliia-livandovskaya (approved):

PR #1639, opened 2023-05-03
Solution

emxm:
https://emxm.github.io/react_tabs/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,40 +1,46 @@
-import React from 'react';
+import { FC, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
+import { Tabs } from './components/Tabs';
 
-export const tabs = [
+export const tabs: Tab[] = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
   { id: 'tab-2', title: 'Tab 2', content: 'Some text 2' },
   { id: 'tab-3', title: 'Tab 3', content: 'Some text 3' },
 ];
 
-export const App: React.FC = () => {
+type Tab = {
+  id: string;
+  title: string;
+  content: string;
+};
+
+export const App: FC = () => {
+  const [selectedTabId, setselectedTabId] = useState('tab-1');
+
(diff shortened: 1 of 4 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,37 @@
...
+  id: string;
+  title: string;
+  content: string;
+};
+
+type Props = {
+  tabs: Tab[],
+  handleChangeTab: (event: React.MouseEvent<HTMLLIElement, MouseEvent>,
+    id: string) => void,
+  selected: string
```
anstsot, line 14: better to rename:
```suggestion
  selectedTab: string
```

src/App.tsx
```diff
@@ -1,40 +1,46 @@
...
+          <Tabs
+            tabs={tabs}
+            handleChangeTab={handleChangeTab}
+            selected={selectedTabId}
+          />
         </div>
 
         <div className="block" data-cy="TabContent">
-          Some text 1
+          {`Some text ${selectedTabId.slice(-1)}`}
```
anstsot, line 43: please use the `content` from selected tab in `tabs` array

src/App.tsx
```diff
@@ -1,40 +1,46 @@
...
 
-export const App: React.FC = () => {
+type Tab = {
+  id: string;
+  title: string;
+  content: string;
+};
+
+export const App: FC = () => {
+  const [selectedTabId, setselectedTabId] = useState('tab-1');
```
anstsot, line 20: don't hardcode the value it's better to use from `tabs` variable
```suggestion
  const [selectedTabId, setselectedTabId] = useState(tabs[0].id);
```

src/App.tsx
```diff
@@ -1,40 +1,46 @@
...
+
+  const handleChangeTab = (event: React.MouseEvent, id: string) => {
+    event.preventDefault();
+    setselectedTabId(id);
```
anstsot, line 24: check if the `selectedId` isn't the same as `id`

anstsot (changes requested):
It’s almost done! Let’s improve a little.
your demo link isn’t working. could you check it, please?

qoa11a (approved):

PR #1647, opened 2023-05-10
react_tabs

Oleksandr-Maslo:
https://oleksandr-maslo.github.io/react_tabs/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,4 +1,5 @@
-import React from 'react';
+import { useState } from 'react';
+import { Tabs, Tab } from './components/Tabs';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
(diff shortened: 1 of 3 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,64 @@
...
+                className={classNames({
+                  'is-active': tab.id === selectedTab.id,
+                })}
+                data-cy="Tab"
+              >
+                <a
+                  href={`#${tab.id}`}
+                  data-cy="TabLink"
+                  onClick={() => {
+                    if (selectedTab !== tab) {
```
AlisaVlasova, line 43: move this if into onTabSelected handler

AlisaVlasova: do not use direct compare on objects, they will never be equal this way

src/App.tsx
```diff
@@ -10,33 +11,19 @@ export const tabs = [
...
-          <ul>
-            <li className="is-active" data-cy="Tab">
-              <a href="#tab-1" data-cy="TabLink">Tab 1</a>
-            </li>
+  const [selectedTabId, setSelectedTabId] = useState('tab-1');
 
-            <li data-cy="Tab">
-              <a href="#tab-2" data-cy="TabLink">Tab 2</a>
-            </li>
+  const onTabSelected = (tab: Tab) => {
```
AlisaVlasova, line 16: you can accept tabId here, you do not use whole tab

AlisaVlasova (changes requested):

sTorba24 (approved):
Well done!

PR #1652, opened 2023-05-13
Done

Ilia991:
https://Ilia991.github.io/react_tabs/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
+import { Tabs } from './components/Tabs';
 
 export const tabs = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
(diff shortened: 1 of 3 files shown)


polosanya (changes requested):
Almost done!

src/App.tsx
```diff
@@ -10,31 +11,36 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [isBtn, setIsBtn] = useState('Tab 1');
```
mykhalenych, line 14: is it for boolean variable
here will be better [currentTab, setCurrentTab]

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,39 @@
...
+      className={
+        classNames(
+          { 'is-active': btn === tab.title },
+        )
+      }
+    >
+      <a
+        href={`#${tab.id}`}
+        data-cy="TabLink"
+        onClick={() => setIsBtn(tab.title)}
```
mykhalenych, line 33: also better to set id instead of text

mykhalenych (approved):

PR #1659, opened 2023-05-18
Solution

NadieinOleh:
https://NadieinOleh.github.io/react_tabs/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
+import { Tabs } from './components/Tabs';
 
 export const tabs = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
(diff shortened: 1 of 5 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,47 @@
...
+        </div>
+
+        <div className="block" data-cy="TabContent">
+          {tabs.map(tab => {
+            if (tab.id === selectedTabId) {
+              return tab.content;
+            }
+
+            return null;
+          })}
```
witflash, line 42: Instead of `map` here you can use `tabs.find` method.

src/App.tsx
```diff
@@ -10,33 +11,27 @@ export const tabs = [
...
-
-      <div data-cy="TabsComponent">
-        <div className="tabs is-boxed">
-          <ul>
-            <li className="is-active" data-cy="Tab">
-              <a href="#tab-1" data-cy="TabLink">Tab 1</a>
-            </li>
+        Selected tab is
+        {' '}
+        {tabs.map(tab => {
```
witflash, line 21: Here you need `tabs.find` method not `map`

OleziO (approved):

PR #1670, opened 2023-05-22
solution

Vaskovics:
<a href="https://vaskovics.github.io/react_tabs/">Demo</a>


src/App.tsx
```diff
@@ -10,33 +11,26 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [selectedTab, setTabId] = useState(tabs[0]);
+  const [myTabs] = useState([...tabs]);
```
SanyaBratashchuk, line 15: This is constant why do you need to hold it in state?

SanyaBratashchuk (changes requested):

OleziO (approved):

PR #1678, opened 2023-05-24
Develop

kacper-lyczba:
[DEMO LINK](https://suaby514.github.io/react_tabs/)

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,4 +23,4 @@ Implement the `App` and `Tabs` components to render and switch between given `ta
 - Implement a solution following the [React task guideline](https://github.com/mate-academy/react_task-guideline#react-tasks-guideline).
 - Use the [React TypeScript cheat sheet](https://mate-academy.github.io/fe-program/js/extra/react-typescript).
 - Open one more terminal and run tests with `npm test` to ensure your solution is correct.
-- Replace `<your_account>` with your Github username in the [DEMO LINK](https://suaby514.github.io/react_tabs/) and add it to the PR description.
+- Replace `<your_account>` with your Github username in the [DEMO LINK](https://kacper-lyczba.github.io/react_tabs/) and add it to the PR description.


src/App.tsx
```diff
@@ -10,33 +11,19 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [selectedTabId, setSelectedTabId] = useState(tabs[0]); // Why does it have to be called selectedTabId? It isn't an id at all
```
d-wasilewski, line 14: Because it should actually be an id
```suggestion
  const [selectedTabId, setSelectedTabId] = useState(tabs[0].id); // Why does it have to be called selectedTabId? It isn't an id at all
```

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,61 @@
...
+  onTabSelected: (tab: Tab) => void,
+  tabs: Tab[],
+}
+
+export const Tabs: React.FC<TabsProps> = ({
+  tabs, selectedTabId, onTabSelected,
+}) => {
+  function findTab(tabId: string): Tab {
+    return tabs.find((tab) => tab.id === tabId) || tabs[0];
+  }
```
d-wasilewski, line 20: assign it to a variable directly instead of creating a function

d-wasilewski (changes requested):

src/App.tsx
SuaBy514: It cannot be an id, when it is an id the Cypress test fails. It specifically requires it to be Object{3}

src/components/Tabs/Tabs.tsx
SuaBy514: A function is created because it is reused at lines 24 and 28

SuaBy514: Unless you mean to use an arrow function instead, which would function virtually the same

src/App.tsx
d-wasilewski: Where exactly?

d-wasilewski: This should be an id, and based on that id you should find a tab with this title and write a function `onTabSelected` that sets the id of a tab instead of an entire tab

d-wasilewski (changes requested):

src/App.tsx
SuaBy514: The test requires the setting function to use the entire tab as an object, which in turn means the state needs to use the tab, and not an id

src/components/Tabs/Tabs.tsx
SuaBy514: Approach with a separate function worked perfectly; I do not know why I did not think of that at all

d-wasilewski (approved):

PR #1702, opened 2023-06-19
addedd tabs component

AlexKherson:
[DEMO LINK](https://alexKherson.github.io/react_tabs/)

### DIFF
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -1,1 +1,79 @@
-export const Tabs = () => {};
+import { FC } from 'react';
+import classNames from 'classnames';
+import { Tab } from '../../types/tab';
+
+type PropsTabs = {
+  tabs: Tab[],
+  selectedTabId: string,
+  onTabSelected: (tab: Tab) => void,
+};
+
+export const Tabs: FC<PropsTabs> = (props) => {
+  const {
+    tabs,
+    selectedTabId,
+    onTabSelected,
+  } = props;
+
+  // useEffect(() => {
+  //   // const isDefaultSelectedId = !tabs
+  //   //   .map(t => t.id)
+  //   //   .includes(selectedTabId);
+
+  //   const isValidTabId = () => {
+  //     return tabs.map(tab => tab.id).includes(selectedTabId);
+  //   };
+
+  //   if (!isValidTabId()) {
+  //     onTabSelected(tabs[0]);
+  //   }
+  // }, []);
+
+  const isValidTabId = () => {
+    return tabs.map(tab => tab.id).includes(selectedTabId);
+  };
+
+  const selected = isValidTabId()
+    ? selectedTabId
+    : tabs[0].id;
+
+  return (
+    <div data-cy="TabsComponent">
+      <div className="tabs is-boxed">
+        <ul>
+          {tabs.map(tab => {
+            const isSelected = selected === tab.id;
+
+            return (
+              <li
+                className={classNames({
+                  'is-active': isSelected,
+                })}
+                data-cy="Tab"
+                key={tab.id}
+              >
+                <a
+                  href={`#${tab.id}`}
+                  data-cy="TabLink"
+                  onClick={() => {
+                    if (isSelected) {
+                      return;
+                    }
+
+                    onTabSelected(tab);
+                  }}
+                >
+                  {tab.title}
+                </a>
+              </li>
+            );
+          })}
+        </ul>
+      </div>
+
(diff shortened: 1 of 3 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,79 @@
...
+  //   //   .includes(selectedTabId);
+
+  //   const isValidTabId = () => {
+  //     return tabs.map(tab => tab.id).includes(selectedTabId);
+  //   };
+
+  //   if (!isValidTabId()) {
+  //     onTabSelected(tabs[0]);
+  //   }
+  // }, []);
```
silchyn, line 30:
```suggestion
```

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,79 @@
...
+
+  const isValidTabId = () => {
+    return tabs.map(tab => tab.id).includes(selectedTabId);
+  };
+
+  const selected = isValidTabId()
```
silchyn, line 36: Selected what?

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,79 @@
...
+  //   const isValidTabId = () => {
+  //     return tabs.map(tab => tab.id).includes(selectedTabId);
+  //   };
+
+  //   if (!isValidTabId()) {
+  //     onTabSelected(tabs[0]);
+  //   }
+  // }, []);
+
+  const isValidTabId = () => {
```
silchyn, line 32: Which tab id?

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,79 @@
...
+                <a
+                  href={`#${tab.id}`}
+                  data-cy="TabLink"
+                  onClick={() => {
+                    if (isSelected) {
+                      return;
+                    }
+
+                    onTabSelected(tab);
+                  }}
```
silchyn, line 64: Put it in a separate callback

silchyn (changes requested):

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,69 @@
...
+
+  const isValidCurrentTabId = () => {
+    return tabs
+      .map(tab => tab.id)
+      .includes(selectedTabId);
+  };
+
+  const currentTabId = isValidCurrentTabId()
+    ? selectedTabId
+    : tabs[0].id;
```
33Orange, line 26: You can use [.find()](https://developer.mozilla.org/ru/docs/Web/JavaScript/Reference/Global_Objects/Array/find) instead
Would be better to find selected tab here to avoid use 1 more find in `TabContent`.

```suggestion
  const selectedTab = tabs.find(tab => tab.id === selectedTabId) || tabs[0];
```

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,69 @@
...
+
+    onTabSelected(tab);
+  };
+
+  return (
+    <div data-cy="TabsComponent">
+      <div className="tabs is-boxed">
+        <ul>
+          {tabs.map(tab => {
+            const isSelected = currentTabId === tab.id;
```
33Orange, line 41:
```suggestion
            const isSelected = selectedTab.id === tab.id;
```

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,69 @@
...
+                  {tab.title}
+                </a>
+              </li>
+            );
+          })}
+        </ul>
+      </div>
+
+      <div className="block" data-cy="TabContent">
+        {tabs.find(tab => tab.id === selectedTabId)?.content}
```
33Orange, line 65:
```suggestion
        {selectedTab.content}
```

33Orange (changes requested):
Almost done

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,62 @@
...
+
+    onTabSelected(tab);
+  };
+
+  return (
+    <div data-cy="TabsComponent">
+      <div className="tabs is-boxed">
+        <ul>
+          {tabs.map(tab => {
+            const isSelected = selectedTab.id === tab.id;
```
dffUqp, line 34: you use `isSelected` only one time, so let's pass this condition straight to classNames object. (remove `return` keyword in this case)

but you used the condition twice in this component, and if you want to leave `isSelected` variable, try to come up with some solution.

dffUqp (approved):
LGTM:fire:

PR #1706, opened 2023-06-20
Develop

serentinos:
[DEMO LINK](https://serentinos.github.io/react_tabs/)

### DIFF
--- a/src/helpers/getTabById.ts
+++ b/src/helpers/getTabById.ts
@@ -1,0 +1,8 @@
+import { Tab } from '../types';
+
+export const getTabById = (
+  tabsList: Tab[],
+  id: string,
+) => (
+  tabsList.find(tab => tab.id === id)
+);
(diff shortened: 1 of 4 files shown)


src/helpers/getTabById.ts
```diff
@@ -0,0 +1,8 @@
+import { Tab } from '../types';
+
+export const getTabById = (
+  tabsList: Tab[],
+  id: string,
+) => (
+  tabsList.find(tab => tab.id === id)
+);
```
dffUqp, line 8: let's handle `undefined` in this function instead of our component.
```suggestion
export const getTabById = (
  tabsList: Tab[],
  id: string,
) => (
  tabsList.find(tab => tab.id === id) || tabsList[0]
);
```

dffUqp (changes requested):
LGTM but needs a few fixes:smiley:

33Orange (approved):
Well done

PR #1709, opened 2023-06-20
solution

okosohor:
[DEMO LINK](https://okosohor.github.io/react_tabs/)

### DIFF
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -14,7 +14,7 @@ export const Tabs: React.FC<Props> = ({
 }) => {
   const isSelectedTab = tabs.find(tab => tab.id === selectedTabId) || tabs[0];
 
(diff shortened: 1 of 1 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,50 @@
...
+};
+
+export const Tabs: React.FC<Props> = ({
+  tabs,
+  selectedTabId,
+  onTabSelected,
+}) => {
+  const isSelectedTab = tabs.find(tab => tab.id === selectedTabId) || tabs[0];
+
+  const HandleCLick = (tab: Tab) => {
```
33Orange, line 17: This is not handle for click. You use this function inside your event handler
```suggestion
  const selectTab = (tab: Tab) => {
```

33Orange (changes requested):
Almost done

Just a one moment with naming

PR #1727, opened 2023-06-21
add solution

ivankhrsh:
[DEMO LINK](https://ivankhrsh.github.io/react_tabs/)

### DIFF
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -1,1 +1,58 @@
-export const Tabs = () => {};
+import cn from 'classnames';
+import { FC } from 'react';
+import { Tab } from '../../types/Tab';
+
+interface Props {
+  tabs: Tab[],
+  selectedTabId: string,
+  onTabSelected: (tab: Tab) => void,
+}
+
+export const Tabs: FC<Props> = ({
+  tabs,
+  selectedTabId,
+  onTabSelected,
+}) => {
+  const selectedTab = tabs.find(tab => tab.id === selectedTabId) || tabs[0];
+
+  const selectTabId = (tab: Tab) => {
+    if (tab.id !== selectedTabId) {
+      onTabSelected(tab);
+    }
+  };
+
+  return (
+    <div className="section">
+      <h1 className="title">
+        {selectedTab && `Selected tab is ${selectedTab.title}`}
+      </h1>
+
+      <div data-cy="TabsComponent">
+        <div className="tabs is-boxed">
+          <ul>
+            {tabs.map(tab => (
+              <li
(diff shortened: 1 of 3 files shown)


dffUqp (changes requested):
LGTM but needs a few fixes:smiley:

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,58 @@
...
+  const selectTabId = (tab: Tab) => {
+    if (tab.id !== selectedTabId) {
+      onTabSelected(tab);
+    }
+  };
+
+  return (
+    <div className="section">
+      <h1 className="title">
+        {selectedTab && `Selected tab is ${selectedTab.title}`}
```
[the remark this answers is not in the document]
ivankhrsh: linter don't like this option
image

is `{`Selected tab is ${selectedTab.title}`}` okay?

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,58 @@
...
+    if (tab.id !== selectedTabId) {
+      onTabSelected(tab);
+    }
+  };
+
+  return (
+    <div className="section">
+      <h1 className="title">
+        {`Selected tab is ${selectedTab.title}`}
+      </h1>
```
33Orange, line 28: Tabs must return only tabs component, but you return full section.

Move this to `app` component and in the `tabs` leave only the `TabsComponent`

33Orange (changes requested):
Almost done

And678 (changes requested):
Your last commit broke tests. 
image
OnTabSelected needs to be called with entire tab data

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,54 @@
...
+export const Tabs: FC<Props> = ({
+  tabs,
+  selectedTabId,
+  onTabSelected,
+}) => {
+  const selectedTab = tabs.find(tab => tab.id === selectedTabId) || tabs[0];
+
+  const selectTabId = (tab: Tab) => {
+    if (tab.id !== selectedTabId) {
+      onTabSelected(tab.id);
```
dffUqp, line 20: let's pass whole `tab` in `onTabSelected`, it will allow us to remove `selectedTab` variable in `App.tsx`. 
(in this case, you need to rename your state to selectedTab and setSelectedTab)

ivankhrsh:
> 



> **[And678](/And678) ** left a comment

There is miss understanding between tests and task in readme

`1. Save the `selectedTabId` in the `App` (the first `tab` is selected by default);` So I save only id

dffUqp (approved):
LGTM:fire:

PR #1732, opened 2023-06-21
add solution

YeMyr:
[DEMO LINK](https://yemyr.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,9 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
+import { Tabs } from './components/Tabs';
+import { Tab } from './Types/Tab';
 
 export const tabs = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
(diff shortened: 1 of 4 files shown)


src/App.tsx
```diff
@@ -10,33 +12,19 @@ export const tabs = [
...
-          </ul>
-        </div>
-
-        <div className="block" data-cy="TabContent">
-          Some text 1
-        </div>
-      </div>
+      <Tabs
+        tabs={tabs}
+        selectedTab={selectedTab}
```
33Orange, line 25: You need to provide only ID of selectedTab to Tabs component
image

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,50 @@
...
+  onTabSelected,
+  selectedTab,
+
+}) => {
+  const selectTab = (tab: Tab) => {
+    if (tab.id !== selectedTab.id) {
+      onTabSelected(tab);
+    }
+  };
+
```
33Orange, line 22: Here you must find `selectedTab` by id from props and if you don't have it use `tabs[0]` as selected tab

33Orange (changes requested):
Almost done

src/components/Tabs/Tabs.tsx
YeMyr: fixed 👌

src/App.tsx
YeMyr: fixed 👌

dffUqp (approved):
LGTM:fire:

PR #1738, opened 2023-06-22
solution

SofiaLiubchenko:
[DEMO LINK](https://SofiaLiubchenko.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,9 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
+import { Tabs, Tab } from './components/Tabs';
+import { findTabById } from './helpers';
 
 export const tabs = [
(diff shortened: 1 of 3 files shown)


And678 (changes requested):
Looks good, one thing to fix, you have dependency cycle here:
`helpers` -> `Tabs` -> `helpers` -> ...

Read [this](https://softwareengineering.stackexchange.com/questions/11856/whats-wrong-with-circular-references) to learn why this is bad.

Anyways, root folder `helpers` shouldn't import stuff from components. Either you can move the function from helpers to components or move your exported `Tab` type to a separate `types.ts`

src/App.tsx
```diff
@@ -10,33 +13,25 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [selectedTabId, setSelectedTabId] = useState<string>('tab-1');
```
33Orange, line 16: What is 'tab-1'?
What would be selected if I have another array of tabs with other values in the property id?

33Orange (changes requested):
Almost done

PR #1759, opened 2023-08-16
add task solution

Luk2asz:
[DEMO LINK](https://Luk2asz.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,8 @@
-import React from 'react';
+import React, { useEffect, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
+import { Tabs } from './components/Tabs';
 
 export const tabs = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -10,33 +11,34 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [selectedTabId, setSelectedTab] = useState('tab-1');
```
darokrk, line 14: You have hard-coded the initially selected tab as 'tab-1'. Consider using the first tab in the 'tabs' array rather than hard-coding to increase flexibility.

`const [selectedTabId, setselectedTabId] = useState(tabs[0]);`

src/App.tsx
```diff
@@ -10,33 +11,34 @@ export const tabs = [
...
+
+  useEffect(() => {
+    if (!tabs.map(tab => tab.id).includes(selectedTabId)) {
```
darokrk, line 17: The use of 'useEffect' here may lead to unnecessary rerenders. Consider moving the check inside the 'onTabSelected' function to only verify when a new tab is selected.

Think you don't need here this useEffect hook.

src/App.tsx
```diff
@@ -10,33 +11,34 @@ export const tabs = [
...
+      setSelectedTab('tab-1');
+    }
+  }, [selectedTabId]);
+
+  const onTabSelected = (newTab: {
```
darokrk, line 22: The 'onTabSelected' function does not check if the newTab is different from the currently selected tab. If it is the same, there is no need to call 'setSelectedTab', saving unnecessary rerenders.

add if check here e.g:

```
if (newTab.id !== selectedTabId) {
      onTabSelected(newTab.id);
}
```

src/App.tsx
```diff
@@ -10,33 +11,34 @@ export const tabs = [
...
+    title: string;
+    content: string }) => {
+    setSelectedTab(newTab.id);
+  };
+
   return (
     <div className="section">
       <h1 className="title">
-        Selected tab is Tab 1
+        Selected tab is
```
darokrk, line 32: this can be writer easier as a template literal string, e.g.

image

src/App.tsx
```diff
@@ -10,33 +11,34 @@ export const tabs = [
...
+        {' '}
+        {tabs.find(tab => tab.id === selectedTabId)?.title || ''}
```
darokrk, line 34: this can be removed now

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,41 @@
-export const Tabs = () => {};
+/* eslint-disable jsx-a11y/click-events-have-key-events */
+/* eslint-disable jsx-a11y/no-noninteractive-element-interactions */
+import '../../App.scss';
+import 'bulma/css/bulma.css';
+import '@fortawesome/fontawesome-free/css/all.css';
+
+type Props = {
+  tabs: { id: string; title: string; content: string; }[];
```
darokrk, line 8: this can be declared above as:

```
interface Tab {
  id: string,
  title: string,
  content: string,
}
```

and use it here directly:

```
type Props = {
  tabs: Tab[],
  ...etc...
}
```

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,41 @@
...
+  selectedTabId: string;
+  onTabSelected: (param: {
```
darokrk, line 10: this can be now:

`onTabSelected: (tab: Tab) => void;`

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,41 @@
...
+          {tabs.map(tab => (
+            <li
+              className={selectedTabId === tab.id ? 'is-active' : ''}
+              data-cy="Tab"
+              key={tab.id}
+              onClick={selectedTabId === tab.id
+                ? () => null
+                : () => onTabSelected(tab)}
+            >
+              <a href={`#${tab.id}`} data-cy="TabLink">{tab.title}</a>
```
darokrk, line 30: think onClick handler should be added directly here to link not list element, please add it here and remove from `<li>` tag:

`onClick={() => handleTabClick(tab)}`

darokrk (changes requested):
Few mention added to improve code 👨🏽‍💻 🛠️

darokrk (approved):
Good job 💪🏽 !

PR #1761, opened 2023-08-29
failing tests 1.0.0

domkab:
https://domkab.github.io/react_tabs/

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,42 +1,37 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
+import { TabInfo } from './utils/types';
+import { Tabs } from './components/Tabs/Tabs';
 
-export const tabs = [
+export const tabs:TabInfo[] = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
   { id: 'tab-2', title: 'Tab 2', content: 'Some text 2' },
   { id: 'tab-3', title: 'Tab 3', content: 'Some text 3' },
 ];
 
 export const App: React.FC = () => {
+  const [activeTabId, setActiveTabId] = useState(tabs[0].id);
+
+  const selectedTab = tabs.find(tab => tab.id === activeTabId);
+
   return (
     <div className="section">
       <h1 className="title">
-        Selected tab is Tab 1
+        {selectedTab ? `Selected tab is ${selectedTab.title}` : 'No tab selected'}
       </h1>
 
-      <div data-cy="TabsComponent">
-        <div className="tabs is-boxed">
-          <ul>
-            <li className="is-active" data-cy="Tab">
-              <a href="#tab-1" data-cy="TabLink">Tab 1</a>
-            </li>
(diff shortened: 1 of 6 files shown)


src/App.tsx
```diff
@@ -1,42 +1,37 @@
...
+export const tabs:TabInfo[] = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
   { id: 'tab-2', title: 'Tab 2', content: 'Some text 2' },
   { id: 'tab-3', title: 'Tab 3', content: 'Some text 3' },
 ];
 
 export const App: React.FC = () => {
+  const [activeTabId, setActiveTabId] = useState(tabs[0].id);
+
+  const selectedTab = tabs.find(tab => tab.id === activeTabId);
```
witflash, line 17: You should default to the first tab if there is no tab:
at the end `|| tabs[0]`

src/App.tsx
```diff
@@ -1,42 +1,37 @@
...
+
   return (
     <div className="section">
       <h1 className="title">
-        Selected tab is Tab 1
+        {selectedTab ? `Selected tab is ${selectedTab.title}` : 'No tab selected'}
```
witflash, line 22: And here you will never have "No tab selected"

src/App.tsx
```diff
@@ -1,42 +1,37 @@
...
-            </li>
-          </ul>
-        </div>
-
-        <div className="block" data-cy="TabContent">
-          Some text 1
-        </div>
+      <div className="block" data-cy="TabContent">
+        {selectedTab ? selectedTab.content : 'No tab selected'}
       </div>
```
witflash, line 33: This content should be inside the `Tabs` component

src/components/Tab/Tab.tsx
```diff
@@ -0,0 +1,24 @@
...
+
+export const Tab: React.FC<TabProps> = ({
+  isActive,
+  title,
+  id,
+  setActiveTabId,
+}) => {
+  return (
+    <li className={isActive ? 'is-active' : ''} data-cy="Tab">
+      <button
```
witflash, line 12: You don't need the `button` here. An anchor inside a button is an error.
Just use `a` tag inside `li` without `button` usage.

witflash (changes requested):
Good work!
Just pay attention to my comments, please. Make all corrections and then check your tests.

src/components/Tab/Tab.tsx
```diff
@@ -0,0 +1,21 @@
...
+  title,
+  id,
+  setActiveTabId,
+}) => {
+  return (
+    <li className={isActive ? 'is-active' : ''} data-cy="Tab">
+      <a
+        href={`#${id}`}
+        data-cy="TabLink"
+        onClick={() => setActiveTabId(id)}
```
Anastasiia-Svintsova, line 15: You are calling setActiveTabId directly in the onClick. It is recommended to make a separate function handleTabClick and call setActiveTabId from that function.

Anastasiia-Svintsova, line 15: image

Anastasiia-Svintsova (changes requested):

lerastarynets (approved):
Keep up the great work!

PR #1806, opened 2023-12-12
Develop

jacekkosiba:
[DEMO LINK](https://jacekkosiba.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,42 +1,28 @@
-import React from 'react';
+import { FC, useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
+import { Tab } from './types';
+import { Tabs } from './components/Tabs';
+import { tabs } from './tabs';
 
-export const tabs = [
-  { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
-  { id: 'tab-2', title: 'Tab 2', content: 'Some text 2' },
-  { id: 'tab-3', title: 'Tab 3', content: 'Some text 3' },
-];
+export const App: FC = () => {
+  const [selectedTabId, setSelectedTabId] = useState('tab-1');
 
-export const App: React.FC = () => {
+  const onTabSelect = (tab: Tab) => {
+    setSelectedTabId(tab.id);
+  };
+
   return (
     <div className="section">
       <h1 className="title">
-        Selected tab is Tab 1
+        {`Selected tab is Tab ${selectedTabId.slice(4)}`}
       </h1>
-
-      <div data-cy="TabsComponent">
-        <div className="tabs is-boxed">
-          <ul>
-            <li className="is-active" data-cy="Tab">
-              <a href="#tab-1" data-cy="TabLink">Tab 1</a>
-            </li>
-
-            <li data-cy="Tab">
-              <a href="#tab-2" data-cy="TabLink">Tab 2</a>
-            </li>
-
-            <li data-cy="Tab">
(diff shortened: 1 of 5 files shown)


kate-prakofyeva (approved):
Good job!

src/App.tsx
```diff
@@ -1,42 +1,32 @@
...
-  { id: 'tab-2', title: 'Tab 2', content: 'Some text 2' },
-  { id: 'tab-3', title: 'Tab 3', content: 'Some text 3' },
-];
+export const App: FC = () => {
+  const [selectedTabId, setSelectedTabId] = useState(tabs[0].id);
+
+  const onTabSelected = (tab: Tab) => {
+    if (tab.id !== selectedTabId) {
+      setSelectedTabId(tab.id);
+    }
```
mvjl000, line 15: I always do it this way. I like to avoid nesting as much as possible.
```suggestion
    if (tab.id === selectedTabId) return;
    
    setSelectedTabId(tab.id);
```

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,36 @@
...
+                key={tab.id}
+                className={tab.id === selectedTab?.id ? 'is-active' : ''}
+                data-cy="Tab"
+              >
+                <a
+                  href={`#${tab.id}`}
+                  data-cy="TabLink"
+                  onClick={
+                    () => tab.id !== selectedTab.id && onTabSelected(tab)
+                  }
```
mvjl000, line 22: Create a handler function. Cleaner jsx

mvjl000 (changes requested):
Well done! Just one small adjustment

src/App.tsx
jacekkosiba: that unfortunately I can't do, linter won't let me, but I can do this:
`  const onTabSelected = (tab: Tab) => {
    if (tab.id === selectedTabId) {
      return;
    }

    setSelectedTabId(tab.id);
  };`

mvjl000 (approved):
Good good

src/App.tsx
```diff
@@ -1,42 +1,34 @@
...
+
+  const onTabSelected = (tab: Tab) => {
+    if (tab.id === selectedTabId) {
+      return;
+    }
+
+    setSelectedTabId(tab.id);
+  };
+
+  const selectedTab = tabs.find(tab => tab.id === selectedTabId);
```
mbruhler, line 20: If you get the whole object here, pass it to the child component, then you won't need to pass the ID.

I mean, you are looking for selectedTab in tabs twice, once here and second time in Tab component. You don't need to, you can do it just here, and pass it as a prop

mbruhler (changes requested):

src/App.tsx
jacekkosiba: This is actually what I intended to do in the beginning of solving this task, but unfortunately the task description says:

> 1. Save the `selectedTabId` in the `App` (the first `tab` is selected by default);
and
> 1. Pass the `selectedTabId` as a prop to the `Tabs`

I think it may be mandatory to do it this way for the tests to go through, because when I am trying to pass the whole selectedTab object as a props, instead of just suing selectedTabId the tests would not pass.

kate-prakofyeva: The same, because at this moment we do not yet have a changed state, so we have to do this check and calculate the current table again 

instead of passing the whole tab

jacekkosiba: I found a workaround by moving all the JSX from the App component to the Tab component. It might not be the most elegant solution, but it functions properly; the tests pass. Now there's no need to pass the entire object as a prop or to check for the selectedTab in the App component.

JoyCoffeeAddict (approved):

PR #1809, opened 2023-12-14
Solution.

adrian200002:
[DEMO LINK](https://adrian200002.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,8 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
+import { Tabs } from './components/Tabs';
 
 export const tabs = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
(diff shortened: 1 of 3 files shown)


src/App.tsx
```diff
@@ -1,42 +1,41 @@
...
 
+ type Tab = {
+   id: string,
+   title: string,
+   content: string
+ };
+
 export const App: React.FC = () => {
+  const [selectedTabId, setSelectedTabId] = useState('tab-1');
+  const [selectedTabTitle, setSelectedTabTitle] = useState('Tab 1');
```
aureliagorzna, line 21: Inspite of tests being poorly designed, the more efficient way to manage the tab properties would be to use just one state for the whole object

src/App.tsx
```diff
@@ -1,42 +1,41 @@
...
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
   { id: 'tab-2', title: 'Tab 2', content: 'Some text 2' },
   { id: 'tab-3', title: 'Tab 3', content: 'Some text 3' },
 ];
 
+ type Tab = {
+   id: string,
+   title: string,
+   content: string
+ };
```
aureliagorzna, line 17: You should practice the good coding habits by always separating the types into the types folder or file

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,54 @@
-export const Tabs = () => {};
+ type Tab = {
+   id: string,
+   title: string,
+   content: string
+ };
+
```
aureliagorzna, line 6: Same here. Props type is an exception because it is used only one time to define the type of the component's props. Consider using interfaces for object types because they can be extended, which is quite common for objects. I think it is a better long term solution.

aureliagorzna (changes requested):

aureliagorzna (approved):

mbruhler (approved):
Yeah @aureliagorzna is right, but this is also a good solution and it will pass tests (despite I wrote an issue for them because I don't like them)

PR #1816, opened 2023-12-21
Solution

ArtemVelychko:
[DEMO LINK](https://ArtemVelychko.github.io/react_tabs/)

### DIFF
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -1,1 +1,59 @@
-export const Tabs = () => {};
+import cn from 'classnames';
+import { TabsType } from '../../types/TabsType';
+
+type Props = {
+  tabs: TabsType[];
+  selectedTabId: string;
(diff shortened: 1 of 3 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,59 @@
...
+                )
+              );
+            })}
+          </ul>
+        </div>
+
+        <div className="block" data-cy="TabContent">
+          {tabs
+            .filter((tab) => tab.id === selectedTabId)
+            .map(({ content }) => (content))}
```
witflash, line 55: I think, you don't need to do this. You already have `selectedTab` and can use it here:
```
<div>
  {selectedTab.content}
</div>
```

witflash (changes requested):
Great work!
I have only one recommendation. Chech my comment and make changes.

lerastarynets (approved):
You're on the right track!

PR #1819, opened 2024-03-04
Solution

wsadocha:
[DEMO LINK](https://wsadocha.github.io/react_tabs/)

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -1,7 +1,9 @@
-import React from 'react';
+import React, { useState } from 'react';
 import 'bulma/css/bulma.css';
 import '@fortawesome/fontawesome-free/css/all.css';
 import './App.scss';
+import { Tabs } from './components/Tabs';
+import { Tab } from './types/Tab';
 
 export const tabs = [
   { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
@@ -10,37 +12,23 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
+  const [selectedTabId, setSelectedTabId] = useState(tabs[0].id);
+
+  const selectedTab = tabs.find(tab => tab.id === selectedTabId) || tabs[0];
+
+  const handleChange = (tab: Tab) => {
+    setSelectedTabId(tab.id);
+  };
+
   return (
     <div className="section">
-      <h1 className="title">Selected tab is Tab 1</h1>
+      <h1 className="title">Selected tab is {selectedTab.title}</h1>
 
-      <div data-cy="TabsComponent">
-        <div className="tabs is-boxed">
-          <ul>
-            <li className="is-active" data-cy="Tab">
-              <a href="#tab-1" data-cy="TabLink">
-                Tab 1
-              </a>
-            </li>
-
-            <li data-cy="Tab">
-              <a href="#tab-2" data-cy="TabLink">
-                Tab 2
-              </a>
-            </li>
-
-            <li data-cy="Tab">
-              <a href="#tab-3" data-cy="TabLink">
-                Tab 3
-              </a>
-            </li>
-          </ul>
-        </div>
-
-        <div className="block" data-cy="TabContent">
-          Some text 1
-        </div>
-      </div>
+      <Tabs
+        tabs={tabs}
+        selectedTabId={selectedTabId}
+        onTabSelected={handleChange}
+      />
     </div>
   );
 };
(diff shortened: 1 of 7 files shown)


yevhenii-pyl (changes requested):
Good job, but let's simplify things here:

src/App.tsx
```diff
@@ -10,37 +12,23 @@ export const tabs = [
 ];
 
 export const App: React.FC = () => {
-  return (
-    <div className="section">
-      <h1 className="title">Selected tab is Tab 1</h1>
+  const [selectedTabId, setSelectedTabId] = useState(tabs[0].id);
```
yevhenii-pyl, line 15: Use just tab, no need to pass the ID and then bother with finders. 

 `const [selectedTab, setSelectedTab] = useState(tabs[0]);`

src/App.tsx
```diff
@@ -10,37 +12,23 @@ export const tabs = [
...
 
-      <div data-cy="TabsComponent">
-        <div className="tabs is-boxed">
-          <ul>
-            <li className="is-active" data-cy="Tab">
-              <a href="#tab-1" data-cy="TabLink">
-                Tab 1
-              </a>
-            </li>
+  const selectedTab = tabs.find(tab => tab.id === selectedTabId) || tabs[0];
```
yevhenii-pyl, line 17: remove, no need now

src/App.tsx
```diff
@@ -10,37 +12,23 @@ export const tabs = [
...
 
-            <li data-cy="Tab">
-              <a href="#tab-2" data-cy="TabLink">
-                Tab 2
-              </a>
-            </li>
+  const handleChange = (tab: Tab) => {
```
yevhenii-pyl, line 19: You are creating a handler (which is generally a good idea), but then there is some logic at the 'onClick' function. 
If you create handlers => handlers should cover ALL logic. 
In there case honestly there is no need to create a handler, just use `setSelectedTab` instead.

src/App.tsx
```diff
@@ -10,37 +12,23 @@ export const tabs = [
...
+    <div className="section">
+      <h1 className="title">Selected tab is {selectedTab.title}</h1>
 
-        <div className="block" data-cy="TabContent">
-          Some text 1
-        </div>
-      </div>
+      <Tabs
+        tabs={tabs}
+        selectedTabId={selectedTabId}
```
yevhenii-pyl, line 29: just selectedTab

src/App.tsx
```diff
@@ -10,37 +12,23 @@ export const tabs = [
...
+        onTabSelected={handleChange}
```
yevhenii-pyl, line 30: just setSelectedTab

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,48 @@
...
+import { Tab } from '../../types/Tab';
+
+type Props = {
+  tabs: Tab[];
+  selectedTabId: string;
+  onTabSelected: (tab: Tab) => void;
+};
+
+export const Tabs = ({ tabs, selectedTabId, onTabSelected }: Props) => {
+  const selectedTab = tabs.find(tab => tab.id === selectedTabId) || tabs[0];
```
yevhenii-pyl, line 11: Well, you have already found a selectedTab at the `App`, you should have passed it as a prop. 
After adjustments there will no need at all at finding a tab by id.

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,48 @@
...
+
+  return (
+    <div data-cy="TabsComponent">
+      <div className="tabs is-boxed">
+        <ul>
+          {tabs.map(tab => {
+            return (
+              <li
+                key={tab.id}
+                className={cn({ 'is-active': selectedTab === tab })}
```
yevhenii-pyl, line 21: Really risky idea to compare objects like that, it may lead to really unwanted behaviour. 
You should compare ID's here

yevhenii-pyl (approved):
Ahh, actually, my bad: suggested code changes break some tests functionality. 

Looks like it's necessary to have that check for a selected tab to check if we have a valid id: 
`const selectedTab = tabs.find(tab => tab.id === selectedTabId) || tabs[0];`
and to operate with `selectedTabId`

Which actually makes it a little less readable and compact, but will pass the tests. 
Since I guided you in the wrong direction with suggested fixes - see no reason to block the PR further, you definitely got the idea. However, it will be great if you fix tests and upload changes.

PR #1822, opened 2024-03-17
add task solution

AgnieszkaWw:
[DEMO LINK](https://AgnieszkaWw.github.io/react_tabs/) and add it to the PR description.

### DIFF
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -1,1 +1,49 @@
-export const Tabs = () => {};
+import classNames from 'classnames';
+import { Tab } from '../../types/Tab';
+
+type Props = {
+  tabs: Tab[];
+  selectedTabId: string;
+  onTabSelected: (tab: Tab) => void;
+};
+
+export const Tabs = ({ tabs, selectedTabId, onTabSelected }: Props) => {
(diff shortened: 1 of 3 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,49 @@
...
+import { Tab } from '../../types/Tab';
+
+type Props = {
+  tabs: Tab[];
+  selectedTabId: string;
+  onTabSelected: (tab: Tab) => void;
+};
+
+export const Tabs = ({ tabs, selectedTabId, onTabSelected }: Props) => {
+  const validTab = tabs.find(tab => tab.id === selectedTabId) || tabs[0];
```
yevhenii-pyl, line 11: Technically, it's a valid tab, however, what if we call it current tab or just selectedTab? 
This may be more accurate.

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,49 @@
...
+      onTabSelected(tab);
+    }
+  };
+
+  return (
+    <div data-cy="TabsComponent">
+      <div className="tabs is-boxed">
+        <ul>
+          {tabs.map(tab => {
+            const { id, title } = tab;
```
yevhenii-pyl, line 23: 👍

PR #1831, opened 2024-04-10
test

marinatea:
[DEMO LINK](https://marinatea.github.io/react_tabs/)


choeqq (changes requested):
please make sure your demo is working

marinatea:
I don't understand, why it doesn't work. There is all seems correct.
When I make npm start - all good too.

marinatea:
I checked by the phone, and the DEMO works good. I do not understand what the problem could be

choeqq (approved):

PR #1833, opened 2024-04-10
Solution

etuzimek:
[DEMO LINK](https://etuzimek.github.io/react_tabs/)


nataliaklonowska (changes requested):
you need to check if selected tab id is the one of given tabs and if it's incorrect assign first tab

nataliaklonowska (approved):

PR #1838, opened 2024-04-12
Develop

spencerspe:
Deploy failed with this error: [eslint] 
src/components/Tabs/index.ts
  Line 1:1:  Dependency cycle via ../../App:1  import/no-cycle

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -3,12 +3,7 @@ import "bulma/css/bulma.css";
 import "@fortawesome/fontawesome-free/css/all.css";
 import "./App.scss";
 import { Tabs } from "./components/Tabs";
-
-export const tabs = [
-  { id: 'tab-1', title: 'Tab 1', content: 'Some text 1' },
-  { id: 'tab-2', title: 'Tab 2', content: 'Some text 2' },
-  { id: 'tab-3', title: 'Tab 3', content: 'Some text 3' },
-];
+import { tabs } from "./components/tabsData"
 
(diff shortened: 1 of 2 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,38 @@
...
+            data-cy="Tab"
+            key={`#${tab.id}`}
+          >
+            <a
+              data-cy="TabLink"
+              href={`#${tab.id}`}
+              onClick={(evt) => {
+                evt.preventDefault();
+                onChange(tab.title, tab.content);
+              }}
```
lerastarynets, line 29: Pls move the function inside `onClick` handler to a separate const and use as a callback here

src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,38 @@
...
+
+type Props = {
+  tabs: Tab[];
+  onChange: (title: string, content: string) => void;
+  activeTab: string;
+};
+
+export const Tabs: React.FC<Props> = ({ tabs, onChange, activeTab }) => {
+  return (
+    <>
```
lerastarynets, line 15: redundant fragment

lerastarynets (changes requested):
Pls refactor your code so that it satisfies the no-cycle rule
[see how to manage it](https://github.com/import-js/eslint-plugin-import/blob/main/docs/rules/no-cycle.md)

PR #1843, opened 2024-05-23
task solved

meljaszuk:
[DEMO LINK](https://meljaszuk.github.io/react_tabs/)


yevhenii-pyl (approved):
There is no magic in React if you've properly learned JS and so many things can be done way easier, isn't it?

Keep going with the good work! 💪

PR #1844, opened 2024-06-10
Develop

aMtiIV:
[DEMO LINK](https://aMtiIV.github.io/react_tabs/)

### DIFF
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -1,1 +1,45 @@
-export const Tabs = () => {};
+import React from 'react';
+import { Tab } from '../../types/Tab';
+
+type Props = {
+  tabs: Tab[];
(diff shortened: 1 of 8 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,45 @@
...
+  }
+
+  return (
+    <>
+      <div className="tabs is-boxed">
+        <ul>
+          {tabs.map(tab => (
+            <li
+              key={tab.id}
+              className={tab.id === selectedTab?.id ? 'is-active' : ''}
```
BogdanMaliuta, line 24: do not need to use here ternary operator with empty option, you can make it like this classname: condition

BogdanMaliuta (approved):
GJ!
Bu next time use CN library in different way

PR #1847, opened 2024-07-16
My solution

Karanelus:
[DEMO LINK](https://karanelus.github.io/react_tabs/)

### DIFF
--- a/src/components/Tabs/Tabs.tsx
+++ b/src/components/Tabs/Tabs.tsx
@@ -1,1 +1,46 @@
-export const Tabs = () => {};
+import { FC } from 'react';
+import { Tab } from '../../types/tab';
+
+type Props = {
+  tabs: Tab[];
+  selectedTabId: string;
+  onTabSelected: (tab: Tab) => void;
+};
+
+export const Tabs: FC<Props> = ({ tabs, selectedTabId, onTabSelected }) => {
+  const selected = tabs.find(tab => tab.id === selectedTabId) || tabs[0];
+
+  const handleClick = (tab: Tab) => {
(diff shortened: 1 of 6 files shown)


src/components/Tabs/Tabs.tsx
```diff
@@ -1 +1,46 @@
...
+    }
+  };
+
+  return (
+    <div data-cy="TabsComponent">
+      <div className="tabs is-boxed">
+        <ul>
+          {tabs.map(tab => (
+            <li
+              className={selected.id === tab.id ? 'is-active' : ''}
```
natalia-klonowska, line 25: it's better to use classnames especially when one of your option in ternary operator is empty

natalia-klonowska (changes requested):

src/components/Tabs/Tabs.tsx
```diff
@@ -1,35 +1,32 @@
...
+                'is-active': selected?.id === tab.id,
+              })}
               data-cy="Tab"
               key={tab.id}
             >
               <a
                 href={`#${tab.id}`}
                 data-cy="TabLink"
-                onClick={() => handleClick(tab)}
+                onClick={() => onTabSelected(tab)}
```
natalia-klonowska, line 29: > After fixes test "should not invoke onTabSelected if the current tab is clicked" not passing, even when worked like need to be

it's because now at onClick you have onTabSelected instead of handleClick ;p change back to handleClick and condition to invoke onTabSelected also should be there again

natalia-klonowska (changes requested):

natalia-klonowska (approved):

